PR #12, opened 2015-05-28
Stop telling users to run `sudo brew`

jacknagel:
It's not necessary and will ultimately cause users grief (see https://github.com/Homebrew/homebrew/issues/39701). We don't support it.

Also, libusb will be installed as a dependency of dfu-util, so users don't need to do so manually.

I'm not interested in signing a CLA, so please feel free to apply these changes manually.


jacknagel:
Copied from https://github.com/spark/spark-cli/pull/193

dmiddlecamp:
Awesome, Thanks! :)

jacknagel:
Thanks!

PR #24, opened 2015-06-08
Webhook fixes

kennethlimcp:
- updated to `deviceID`
- print out `Hook ID` upon successful creation
- removed redundant `if (eventName && !url && !coreID)` as it is never
executed due to the `if statement` above
- check that file extension is a `.json` before parsing
- add prompt for empty `deviceID` since that’s a require param

Output:

- `particle webhook create my_event my_url my_id`

```bash
Successfully created webhook with id: 55752ff469928dd61748b0a0
```


kennethlimcp:
Gonna clean up and send a new PR!

PR #25, opened 2015-06-08
Remove # from webhook list

kennethlimcp:
From: https://github.com/spark/spark-cli/pull/197


emilyrose:
definitely less confusing, thank you!

PR #26, opened 2015-06-08
Update Readme.md

kennethlimcp:
- change `core` to `device`
- CAPS —> `DFU` and `Core`
- use `deviceName` instead of `device_id`
- standardize use of “```” for code formatting


emilyrose:
hey, thanks for all of this, looks great! :smile_cat:

PR #27, opened 2015-06-08
Print-out account username

kennethlimcp:
- This should reduce @_@ by showing which account the user is currently on.
- Mention `enter password` instead of `re-enter` since they are not re-entering

Fixes: https://github.com/spark/spark-cli/pull/177


emilyrose:
this is a great idea, thanks for adding clarity here!

PR #28, opened 2015-06-08
add `.der` to saved keys file

kennethlimcp:
- core —> device
- require user to only specify a filename as mentioned in `console.log()`

From: https://github.com/spark/spark-cli/pull/176


emilyrose:
looks great; thank you again! :smile_cat: \

PR #30, opened 2015-06-08
resolve error if renaming fails

kennethlimcp:
- core —> device
- console output is now removed from `ApiClient.js`

Old output:

```sh
Failed to rename device, server said  [ 'Name already in use' ]
Potentially unhandled rejection [1] {"ok":false,"errors":["Name already
in use"]} (WARNING: non-Error used)
```

Now:

```sh
Failed to rename photon-1, server said [ 'Name already in use' ]
```


emilyrose:
you are a gentlehacker and a scholar!

PR #31, opened 2015-06-08
rename core to device

kennethlimcp:
Did this manually so it shouldn’t be breaking any functionality.

- Spark —> Particle
- coreid/coreId —> deviceID
- cores —> devices
- core —> device


emilyrose:
I feel like we've already got this fixed, yeah? Correct me if I'm wrong!

PR #34, opened 2015-06-09
Updated URL for DFU Utils

KatieK2:
It looks like the DFU Utils are no longer hosted at GNUMonks? I've replaced those links with the one from Source Forge.


emilyrose:
thanks, @KatieK2 :smile_cat: very helpful!

PR #35, opened 2015-06-09
Webhook fixes

kennethlimcp:
- update `.json` template
- coreID —> deviceid
- check that `.json` file is found
- check that `.json` file can be parsed or fail
- `particle webhook GET` works now
- `particle webhook POST` works now
- `particle webhook create xxx.json` has also been tested
- print out `hookID` upon hook creation


kennethlimcp:
Bump: https://community.particle.io/t/breezospark-air-pollution-station/11270/31

PR #39, opened 2015-06-12
fixed resolve error

kennethlimcp:
Based on https://github.com/spark/particle-cli/commit/31299a53098c6336be1e35e1db77367e4b1f8361,
the change from `for loop()` to `forEach` does not bail out on the `return` and ends up executing code below.

Without this fix, there's always a `Apparently I didn't find a DFU device? util said  dfu-util 0.8` statement


emilyrose:
great catch. thanks again! :100:

PR #40, opened 2015-06-14
Compile for specific device

kennethlimcp:
- take in extra param for `deviceType`
- Fixes: https://github.com/spark/particle-cli/issues/15
- Fixes: https://github.com/spark/particle-cli/issues/16

**Docs:**

To compile code using the Cloud build farm, you need to specify the target device. 

- eg. Target a Photon
    - `particle compile photon xxxx`
    - `particle compile p xxxx` <---- short-form to reduce typing per compile

- eg. Target a Core
    - `particle compile core xxxx`
    - `particle compile c xxxx` <---- short-form to reduce typing per compile

- eg. No device specified 
    - `particle compile`

    Error:  `Please specify the target device type. eg. particle compile photon xxx`

The current PR insists on a deviceType to be passed but we can also make it optional and default to `core` if that's better.

### DIFF
--- a/commands/CloudCommands.js
+++ b/commands/CloudCommands.js
@@ -259,10 +259,36 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
 
 	},
 
-	compileCode: function () {
-		//  "Please specify a binary file, source file, or source directory");
+	compileCode: function (deviceType) {
+		if (!deviceType) {
+			console.error("\nPlease specify the target device type. eg. particle compile photon xxx\n");
+			return;
+		}
 
-		var args = Array.prototype.slice.call(arguments, 0);
+		//defaults to 0 for core
+		var platform_id = 0;
+
+		switch (deviceType) {
+		  case "photon":
+		  case "p":
+				deviceType = "photon";
+				platform_id = 6;
+		    break;
+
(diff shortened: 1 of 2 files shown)


loopj:
This looks great, would love this to be merged

emilyrose:
Hey, this looks great. Thanks for the PR, and my apologies for the delay in merging!

ScruffR:
The new parameter seems to be compulsory as I tried an old batch which only called `particle compile .`  build the from current location against the Core, but then I got 
```
Target device . not valid
```
But doing `particle compile help` didn't tell anything about the new parameter.

loopj:
The new usage is:

`particle compile photon .` or `particle compile spark .`

ScruffR:
I'm aware of that, but my main point was that `particle help compile` doesn't reflect this.

And since `core` is not the default target (which would have rendered `help` at least somewhat correct for legacy users) it's outright wrong now.

emilyrose:
If this is causing some confusion I'm happy to adjust the logic here.

Perhaps we make it prompt you if you don't specify a device type? (this would have implications for using the CLI as part of a build script)

We can also just default to the `core`, or allow you to specify your own default via a config setting.

PR #42, opened 2015-06-18
Fix URL for listening mode

objectiveSee:


emilyrose:
thanks again! I'm glad you caught these. :smile:

PR #43, opened 2015-06-18
Fix path for documentation URLs

objectiveSee:
Looks like "core" needs to be lowercase in the path.


emilyrose:
oh hey @objectiveSee, great catch. thank you! :star:

PR #64, opened 2015-07-12
validate SSID prompt

kennethlimcp:
This will check that the non-empty SSID is entered before prompting further.

Fixes: https://github.com/spark/particle-cli/issues/62

Error message is: `>> please enter a valid SSID`


emilyrose:
:+1: thanks!

Issue #51, opened 2015-06-25
Incomplete use of 'sudo' when running 'dfu-util'


endorama:
In my setup I need to use `sudo` to properly run `dfu-util -l` ( the setup has been created following the [guide provided](https://community.particle.io/t/how-to-install-spark-cli-on-ubuntu-12-04/3474) in the Readme of this repository ).

There is a configuration option ( `useSudoForDfu` ) that I presume should be used in this cases, but currently  doesn't run `dfu-util -l` with `sudo` ( actually is used only for **reading** from the device, not for listing or writing ), as you can see from the lines in `lib/dfu.js`:
* [lib/dfu.js line 44](https://github.com/spark/particle-cli/blob/88ea621d2ea64aa85c112efc4d0a91a6d6634799/lib/dfu.js#L44)
* [lib/dfu.js line 87](https://github.com/spark/particle-cli/blob/88ea621d2ea64aa85c112efc4d0a91a6d6634799/lib/dfu.js#L87)
* [lib/dfu.js line 111](https://github.com/spark/particle-cli/blob/88ea621d2ea64aa85c112efc4d0a91a6d6634799/lib/dfu.js#L111)
* [lib/dfu.js line 93](https://github.com/spark/particle-cli/blob/88ea621d2ea64aa85c112efc4d0a91a6d6634799/lib/dfu.js#L93) where `getCommandPrefix` is run and returns `sudo dfu-util` if `useSudoForDfu` is true

This makes the commands like `keys` crash ( in particular I wasn't able to complete step 7 of spark-server installation process ).

Is this the intended behaviour? I tried changing the code to use sudo and everything seemed to work fine.

I can submit a PR to fix this.
Thank you

kennethlimcp:
@endorama , do you have a PR? Or i can fix this :)

endorama:
PR submitted! :)

emilyrose:
yay! thanks @endorama! :heart:

PR #65, opened 2015-07-13
Fix for Incomplete use of 'sudo' when running 'dfu-util'

endorama:
Use sudo when `useSudoForDfu` option is `true` for every dfu command.

Should fix issue #51

### DIFF
--- a/lib/dfu.js
+++ b/lib/dfu.js
@@ -41,7 +41,8 @@ var that = module.exports = {
 		var temp = when.defer();
 
 		var failTimer = utilities.timeoutGenerator("findCompatibleDFU timed out", temp, 6000);
-		child_process.exec("dfu-util -l", function (error, stdout, stderr) {
+		var cmd = that.getCommand() + " -l";
+		child_process.exec(cmd, function (error, stdout, stderr) {
 			clearTimeout(failTimer);
 			Object.keys(specs).forEach(function (id) {
 				if(stdout.indexOf(id) >= 0) {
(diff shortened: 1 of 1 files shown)


lib/dfu.js
```diff
@@ -108,12 +109,16 @@ var that = module.exports = {
...
 
 		that.checkBinaryAlignment("-D " + binaryPath);
-		return utilities.deferredSpawnProcess("dfu-util", args);
+		return utilities.deferredChildProcess(that.getCommand() + ' ' + args.join(' '));
 	},
 
-	getCommandPrefix: function () {
-		var sudo = (settings.useSudoForDfu) ? "sudo " : "";
-		return sudo + "dfu-util -d " + that.deviceID;
+	getCommand: function () {
```
kennethlimcp: @emilyrose  i'm thinking this should have a more specific name. Thoughts?

kennethlimcp:
@endorama, the three commits can be squashed into one commit. Using `git rebase -i`, you change `pick` for the older two commits to `squash` and `git push` then they will be combined into one.

lib/dfu.js
endorama: Just to clarify on this name: I decided to use `getCommand()` because this method is located in dfu module ( and actually permit to get the proper dfu command to run ).

endorama:
@kennethlimcp squashed!

kennethlimcp:
LGTM :). Will see what emily says :)

emilyrose:
hey, looks good! thanks @endorama! thanks for your help, @kennethlimcp :smile_cat: :+1:

lib/dfu.js
```diff
@@ -108,12 +109,16 @@ var that = module.exports = {
 		];
 
 		that.checkBinaryAlignment("-D " + binaryPath);
-		return utilities.deferredSpawnProcess("dfu-util", args);
+		return utilities.deferredChildProcess(that.getCommand() + ' ' + args.join(' '));
```
kennethlimcp: @endorama, is there some reason this need to be `deferredChildProcess()` as it broke the verbose display for DFU.

endorama: Using `deferredSpawnProcess` produce a `ENOENT` error ( there is a reference for this in the long commit message for #65 ).
I do not remember precisely why it worked creating a child process instead of spawning a new one.

PR #79, opened 2015-08-03
Fix knownApps flashing issue

kennethlimcp:
Issue is from the unresolved path for the binary location.

- Path fixing function moved to settings.js
- remove "binaries/" path from specfications.js to prevent path checking issues from slipping through.
- clean up deviceSpecs index.js for now


emilyrose:
thanks again, @kennethlimcp :heart:

PR #82, opened 2015-08-03
Prompt to install UDEV rules for dfu

### LINKED ISSUE #74: Udev reference
https://community.particle.io/t/dfu-util-udev-rules/11440

monkbroc:
On Linux only root can access USB devices through dfu. Right now doing `particle flash --usb` will say "no dfu device found" without suggesting how to fix the issue.

It's possible to use sudo to work around this, but a better solution is to install UDEV rules to allow any user to access a particular device.

This PR prompts the user to install UDEV rules when the dfu-util failure indicates missing permissions.

The UDEV rules was checked for the Photon on Ubuntu 15.10. I included rules for the Core, P1 and Electron, but I can't try those.

Somebody working on the CLI should review the code and check if I put the code in the right place as I'm not familiar with the structure of the application.

This fixes #74.

### DIFF
--- a/lib/dfu.js
+++ b/lib/dfu.js
@@ -23,7 +23,7 @@ Lesser General Public License for more details.
 You should have received a copy of the GNU Lesser General Public
 License along with this program; if not, see <http://www.gnu.org/licenses/>.
  ******************************************************************************
- */
+*/
 
 var fs = require('fs');
 var when = require('when');
@@ -45,7 +45,6 @@ var that = module.exports = {
 	deviceID: undefined,
 	findCompatibleDFU: function () {
 		var temp = when.defer();
-		var that = this;
 
 		var failTimer = utilities.timeoutGenerator('findCompatibleDFU timed out', temp, 6000);
 		var cmd = that.getCommand() + ' -l';
@@ -53,7 +52,12 @@ var that = module.exports = {
 			clearTimeout(failTimer);
 			if (error || stderr) {
 				console.error(error || stderr);
-				return temp.reject(error || stderr);
+
+				if(that._missingDevicePermissions(stderr) && that._systemSupportsUdev()) {
+					return that._promptInstallUdevRules().then(temp.reject, temp.reject);
+				} else {
+					return temp.reject(error || stderr);
+				}
 			}
 
 			// find DFU devices that match specs
@@ -293,6 +297,55 @@ var that = module.exports = {
 			segment.specs.address,
 			leave
 		);
+	},
+
+	_udevRulesDir: '/etc/udev/rules.d/',
+
+	_missingDevicePermissions: function (stderr) {
+		return stderr && stderr.indexOf("Cannot open DFU device") >= 0;
+	},
+
+	_systemSupportsUdev: function() {
+		return fs.existsSync(that._udevRulesDir);
+	},
+
+	_promptInstallUdevRules: function() {
+		var temp = when.defer();
+		console.log(chalk.yellow('You are missing the permissions to use DFU without root.'));
+		prompt([{
+			type: 'confirm',
+			name: 'install',
+			message: 'Would you like to install a UDEV rules file to get access?',
+			default: true
+		}], function(ans) {
+			that._installUdevChoice(ans, temp);
+		});
+
+		return temp.promise;
+	},
+
+	_installUdevChoice: function(ans, promise) {
+		var message = 'Missing permissions to use DFU';
+		if (ans.install) {
+			var rules = __dirname + '/50-particle.rules';
+			var cmd = "sudo cp '" + rules + "' '" + that._udevRulesDir + "'";
+			console.log(cmd);
+			child_process.exec(cmd, function(error, stdout, stderr) {
+				if(error) {
+					console.error('Could not install UDEV rules');
+					promise.reject(message);
+				} else {
+					console.log('UDEV rules for DFU installed.');
+					console.log(chalk.bold.red(
+						'Physically unplug and reconnect the Particle device.\n' +
+						'Then run the particle command again.'
+					));
+					promise.resolve(message);
+				}
+			});
+		} else {
+			promise.reject(message);
+		}
 	},
 
 	_: null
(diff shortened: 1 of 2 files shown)


emilyrose:
hey @monkbroc,

This is a great idea! I'd like the chance to chat with some of the folks on the team to get an idea for how we want to approach automagically modifying system configuration for our users. I'll start a conversation about this and follow up here if there's any changes we should make.

Thanks for the PR! I like what you did here.

kennethlimcp:
Fixes: https://github.com/spark/particle-cli/issues/74

kennethlimcp:
I'm thinking that we should check for the **OS** type before doing the addition of udev since that doesn't work on windows (though windows shouldn't have issues with detecting)

monkbroc:
OS check is done by checking if the UDEV rules directory exists before prompting to install UDEV rules.

kennethlimcp:
Hmm interesting but it seems like the attempt is to try copying and if it fails then it's not true. That's not OS checking IMO.

emilyrose:
We can also just use `os.platform()` :smile_cat:

monkbroc:
This line checks if the prompt should be shown:

```
 fs.existsSync(that._udevRulesDir)
```

where `_udevRulesDir = '/etc/udev/rules.d/'`

The idea was that it's better to check for the presence of a feature (UDEV rules folder) than infer that presence from other information (OS name and version). Sure the check can be turned of in Windows and OSX, but what if a user is on a Linux system without UDEV?

kennethlimcp:
So how's yours handling if the user is like what you said - on a Linux system without UDEV?

monkbroc:
I did add printing the stderr in case the prompt isn't shown: `console.log("Apparently I didn't find a DFU device? util said ", stdout, stderr);`

This was another issue here: the user was given no hint that there was an error, only that a device was not found.

monkbroc:
Here's the before behavior: no hint that the issue is missing permissions since `stderr` was swallowed. It seems more like the issue is that device is not in DFU mode.

before udev prompt

Here's the after behavior: prompt, then after a reconnect the flashing works properly.

after udev

monkbroc:
Don't take my comments the wrong way: I'm not against any changes here. I agree that the structure of the code in this PR can be improved.

lib/dfu.js
```diff
@@ -212,5 +223,43 @@ var that = module.exports = {
 		);
 	},
 
+	_udevRulesDir: '/etc/udev/rules.d/',
+
+	_promptInstallUdevRules: function() {
+		var temp = when.defer();
+		console.log(chalk.yellow("You are missing the permissions use dfu without root."));
```
brycekahle: Grammar: `permissions to use`

lib/dfu.js
```diff
@@ -212,5 +223,43 @@ var that = module.exports = {
...
+			var rules = __dirname + "/50-particle.rules";
+			var cmd = "sudo cp \"" + rules + "\" \"" + that._udevRulesDir + "\"";
+			console.log(cmd);
+			child_process.exec(cmd, function(error, stdout, stderr) {
+				if(error) {
+					console.log("Could not install rules: ", stdout, stderr);
+					promise.reject();
+				} else {
+					console.log("UDEV rules for dfu installed.");
+				    console.log(chalk.bold.red("Physically unplug and reconnect the Particle device.\n" +
```
brycekahle: Can you fix the weird indent here?

lib/dfu.js
```diff
@@ -51,8 +53,17 @@ var that = module.exports = {
...
 			});
-			if (that.deviceID == undefined){
-				console.log("Apparently I didn't find a DFU device? util said ", stdout);
+
+			// If the permissions are missing and the system supports
+			// UDEV rules, prompt to install UDEV rules for dfu
+			if(stderr.indexOf("Cannot open DFU device") >= 0 &&
+			  fs.existsSync(that._udevRulesDir)) {
+				console.log(stderr);
+				that._promptInstallUdevRules().then(function () {
```
brycekahle: You are not handling the promise rejection here.

brycekahle:
@monkbroc Thanks for the PR. I left a couple comments. Otherwise it looks good. Hopefully we can get those fixed and get this merged soon.

brycekahle:
It looks like you need to rebase off master too.

monkbroc:
I rebased off master and moved the condition to prompt the UDEV rules install in the error checking block of `child_process.exec`

lib/dfu.js
```diff
@@ -45,7 +45,6 @@ var that = module.exports = {
 	deviceID: undefined,
 	findCompatibleDFU: function () {
 		var temp = when.defer();
-		var that = this;
```
monkbroc: I removed this duplicate `that` since it's already defined a couple lines above. The local `that` shadows the top-level `that`.

Ah, Javascript... https://twitter.com/bhalp1/status/578925947245633536

lib/dfu.js
```diff
@@ -295,5 +299,54 @@ var that = module.exports = {
 		);
 	},
 
+	_udevRulesDir: '/etc/udev/rules.d/',
+
+	_missingDevicePermissions: function (stderr) {
+		return stderr && stderr.indexOf("Cannot open DFU device") >= 0;
```
brycekahle: Can this error happen if the device is simply not plugged in? We probably need to check to see if the rules are already installed.

monkbroc: No, the full error is `dfu-util: Cannot open DFU device 2b04:d006` so it only happens when a device is plugged in but the file permissions won't allow opening it.

brycekahle: Ok. Can you add a check for the case when the rules are installed but there is still a permissions issue? We shouldn't prompt to install the rules in this case.

monkbroc: I'd go with no because I don't know when that situation would happen. If the rules are installed and the permissions are still denied, what do you say to the user? Sorry?

brycekahle: We can show an error message, but we shouldn't try to install the rules again. That won't fix anything.

monkbroc:
I added the check for the rules already existing.

brycekahle:
:+1:

PR #86, opened 2015-08-04
Use spawn over child process

kennethlimcp:
Reverting to the original way to call a process until there's sufficient reasons as to why a child process over a spawn is required.

This is related to https://github.com/spark/particle-cli/issues/80 where commit https://github.com/spark/particle-cli/pull/65 essentially removed dfu verbose which is not a deliberate decision.


brycekahle:
Fixed with https://github.com/spark/particle-cli/commit/50572d0be047dd85c67674d36697c6d1d8fc9b6b

PR #87, opened 2015-08-04
Removing device type prompt

kennethlimcp:
There's no need to select the device type you are flashing as the cloud knows it all :)

Comments here: https://github.com/spark/particle-cli/commit/9ae0f127806fb9e65b9c36d67a1796b2044b91e1


brycekahle:
That is only used when flashing a known app. It serves as a guard against attempting to flash a binary that is intended for another device type (e.g. flashing `core_tinker.bin` to a `Photon`)

brycekahle:
@kennethlimcp ok if I close this?

kennethlimcp:
@brycekahle sure

PR #120, opened 2015-09-09
fixes native types being form-encoded which causes them to lose type info

dmiddlecamp:
uses a bearer token and a JSON request so we don't form-encode /  'stringify' native types in the request body.  Fixes https://community.particle.io/t/webhook-adding-double-quotes-to-boolean-value/15199/2

### DIFF
--- a/commands/WebhookCommands.js
+++ b/commands/WebhookCommands.js
@@ -162,7 +162,6 @@ WebhookCommand.prototype = extend(BaseCommand.prototype, {
 		data.event = eventName;
 		data.url = url;
 		data.deviceid = deviceID;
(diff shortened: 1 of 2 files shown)


commands/WebhookCommands.js
```diff
@@ -162,7 +162,7 @@ WebhookCommand.prototype = extend(BaseCommand.prototype, {
 		data.event = eventName;
 		data.url = url;
 		data.deviceid = deviceID;
-		data.access_token = api._access_token;
+		//data.access_token = api._access_token;
```
brycekahle: Can you just delete this line?

dmiddlecamp:
thx

PR #133, opened 2015-09-16
Added --public flag for publish

kennethlimcp:
- defaults to private for publishing
- Fixes: https://github.com/spark/particle-cli/issues/132
- This should not be merged in until i write up some docs and confirm the syntax with @nexxy

### DIFF
--- a/commands/PublishCommand.js
+++ b/commands/PublishCommand.js
@@ -24,7 +24,7 @@ You should have received a copy of the GNU Lesser General Public
 License along with this program; if not, see <http://www.gnu.org/licenses/>.
  ******************************************************************************
  */
- 
+
 var when = require('when');
 var sequence = require('when/sequence');
 
@@ -55,7 +55,8 @@ PublishCommand.prototype = extend(BaseCommand.prototype, {
 	},
 
 
-	publishEvent: function (eventName, data) {
+	publishEvent: function (eventName, data, setPrivate) {
+
 		var api = new ApiClient(settings.apiUrl, settings.access_token);
 		if (!api.ready()) {
 			return -1;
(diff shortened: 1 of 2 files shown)


kennethlimcp:
```

KENMBP:particle-cli kennethlimcp$ node app.js publish my-event
Published private event: my-event

KENMBP:particle-cli kennethlimcp$ node app.js publish my-event --public
Published public event: my-event

KENMBP:particle-cli kennethlimcp$ node app.js publish my-event someData --public 
Published public event: my-event

KENMBP:particle-cli kennethlimcp$ node app.js publish my-event someData
Published private event: my-event

KENMBP:particle-cli kennethlimcp$ 
```

m-mcgowan:
I'd prefer for events to be public by default (like the firmware API) and add a `--private` flag. (Keep `--public` in case people want to be explicit about it to make scripts easier to read.)

kennethlimcp:
@m-mcgowan, i changed it to the `--private` flag that sets it to private and defaults to `private: true` but skipped keeping the `--public` flag everything without `--private` flag will be just set to public. 

Thoughts?

m-mcgowan:
Sounds good, but for me I'm a fan of having an explicit option for every default option. But that's just a matter of taste, so you're free to leave it.

commands/PublishCommand.js
```diff
@@ -66,7 +67,17 @@ PublishCommand.prototype = extend(BaseCommand.prototype, {
 			return -1;
 		}
 
-		api.publishEvent(eventName, data);
+		if(data == "--private" && setPrivate === undefined){
+			setPrivate = true;
+			data = "";
```
brycekahle: This should be `data = undefined;` to maintain backwards compatibility.

brycekahle:
@kennethlimcp Thanks for the PR! Please fix the one issue and then I will merge.

PR #141, opened 2015-09-28
Add ability to compile Electron firmware

suda:


commands/CloudCommands.js
```diff
@@ -210,7 +210,8 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
 					choices: [
 						'Photon',
 						'Core',
-						'P1'
+						'P1',
+						'Electron'
```
brycekahle: Unless we have some "known apps" for the Electron added, there is no point to add it here.

suda: We'll gonna add them anyway so one thing less to remember then. Missing binary is gonna get caught in `prompt` callback so this doesn't cause code to crash.

PR #147, opened 2015-10-04
Ensure that Core, Photon & P1 are able to flash voodoospark

rwaldron:


emilyrose:
oh, thank you! :smile_cat:

brycekahle:
They can't all use the same binary though correct?

Issue #163, opened 2015-10-27
particle flash does not return failure code


fossum:
When running a build integration tool, the exit code are important, because they determine if the build succeeded. `particle flash` doesn't seem to use the correct codes though. This is the output from my build server:

```sh
$ particle flash fridge_sensor main.ino
Including:
    main.ino
attempting to flash firmware to your device fridge_sensor
Flash device failed
[object Object]

Build succeeded.
```

And here are a couple commands from my development computer.

```sh
ericfoss@UBUNTU64:~/particle/smart-fridge$ particle flash fridge_sensor main.ino
Including:
    main.ino
attempting to flash firmware to your device fridge_sensor
Flash device failed
[object Object]
ericfoss@UBUNTU64:~/particle/smart-fridge$ echo $?
0
ericfoss@UBUNTU64:~/particle/smart-fridge$
```

Above seems to be from "commands/CloudCommands.js", but below seems to be in a lot of different places.

```sh
[ericfoss@localhost smart-fridge]$ particle flash fridge_sensor main.ino
Including:
    main.ino
You're not logged in. Please login using particle cloud login before using this command
[ericfoss@localhost smart-fridge]$ echo $?
0
[ericfoss@localhost smart-fridge]$ 
```

All of these show an exit code of 0, which is probably wrong...

fossum:
I've written a fix to a branch and filed a pull request #164. As stated there, the fix is only for that function. A global fix should be considered before accepting the changes.

PR #164, opened 2015-10-27
Attempting doFlash exit code fix

fossum:
This change (tested) runs exactly the same as before, but now will exit with an error code. This is not a global fix and only applies to this function.

Please consider a better fix before approving this edit.


brycekahle:
Thanks for the PR. I did pursue a more complete fix for #163 with https://github.com/spark/particle-cli/commit/b44934676d42220c7e0eed1d031e9cd9c82fe430.

I will start going through the other commands soon.

PR #167, opened 2015-11-04
Add Coveralls reporting

suda:


dmiddlecamp:
seems pretty safe?

suda:
Yeah, it doesn't touch how whole CLI works

towynlin:
Just checked out the branch and ran all the same commands travis does, including `npm run coveralls` :+1: Thank you @suda !!

.travis.yml
```diff
@@ -5,8 +5,13 @@ node_js:
 - '4'
 sudo: false
 script:
-- npm run lint
-- npm test
+  - npm run lint
+  - npm test
+matrix:
+  allow_failures:
+    - node_js: "4"
```
brycekahle: This is old, we no longer want to allow failures on node 4.

PR #170, opened 2015-11-20
Update publish docs

kennethlimcp:
- mention about public/private flags
- updated the syntax example

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,7 @@ old_version.cpp
 ```sh
 $ particle compile device_type app.ino library1.cpp library1.h
 ```
+
 #### Compiling in a directory containing project files
 
  This will push all the files in a directory that the command line is currently 'cd' in for compilation.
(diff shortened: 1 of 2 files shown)


kennethlimcp:
@brycekahle there's some weirdness with the tabs as usual but i hope to set it up properly on my Atom IDE to prevent this the next time

README.md
```diff
@@ -413,7 +414,19 @@ $ particle subscribe eventName deviceName
 $ particle subscribe eventName 0123456789ABCDEFGHI
 ```
 
+### particle publish
+
+	Allows a message to be published via the CLI without using a physical Particle device. This is particularly useful when you are testing your firmware against an actual `published` event.
+
+	There is a `--private` flag that allows you to `publish` events to devices subscribing to events with the `MY_DEVICES` option. You can also explicitly use the `--public` flag though that is the default setting.
```
brycekahle: The `--public` flag does not exist, so I don't think we should mention it.

kennethlimcp:
Oh i thought i had that implemented after a discussion with Mat. Will update the docs.

kennethlimcp:
@brycekahle updated the docs nows

brycekahle:
:+1:

PR #175, opened 2015-12-02
Check for update to particle-cli

### LINKED ISSUE #138: Warn user if using an old version
Several issues reported on the forum happen because people use an old version of the CLI. The Heroku CLI checks if the current CLI is the latest version (once an hour at most). See [their implementation](https://github.com/heroku/heroku/blob/2ce77935853d01babc1852a5df2d5848a0b8e668/lib/heroku/updater.rb)

The Particle CLI could check if the current version matches the latest version on GitHub https://api.github.com/repos/spark/particle-cli/releases/latest and advise the user to update

brycekahle:
This stores a timestamp of the last update check and version (if newer)
in the profile.json file.

Fixes #138

screen shot 2015-12-01 at 7 17 23 pm


brycekahle:
@monkbroc thoughts?

lib/update-check.js
```diff
@@ -0,0 +1,65 @@
...
+var semver = require('semver');
+var info = require('../package.json');
+var settings = require('../settings');
+var Spinner = require('cli-spinner').Spinner;
+var chalk = require('chalk');
+Spinner.setDefaultSpinnerString(Spinner.spinners[7]);
+
+function check(next) {
+	console.error('particle-cli v' + info.version);
+	console.error();
```
monkbroc: I wouldn't add this new output at the beginning of every CLI invocation. Is this only for debugging on this branch?

lib/update-check.js
```diff
@@ -0,0 +1,65 @@
...
+
+	if (settings.disableUpdateCheck) {
+		return next();
+	}
+
+	var now = Date.now();
+	var lastCheck = settings.profile_json.last_version_check || 0;
+	// check at most once an hour
+	if (now - lastCheck >= (60 * 60 * 1000)) {
```
monkbroc: This could be longer. 1 day maybe?

lib/update-check.js
```diff
@@ -0,0 +1,65 @@
...
+		return;
+	}
+
+	start(next);
+}
+
+function checkVersion(now, next) {
+	var spin = new Spinner('Checking for updates...');
+	spin.start();
+	latestVersion(info.name).then(function (version) {
```
monkbroc: Can you add a timeout here? You could `race` 2 promises, one that just rejects after a timeout and the one returned by `latestVersion`. I'm thinking what would happen if you're not connected to the internet and you run the CLI.

app.js
```diff
@@ -26,8 +26,12 @@
...
-cli.handle(process.argv, true);
+var updateCheck = require('./lib/update-check');
+
+updateCheck(function () {
+	var Interpreter = require('./lib/interpreter.js');
+	var cli = new Interpreter();
+	cli.supressWarmupMessages = true;
+	cli.startup();
+	cli.handle(process.argv, true);
+});
```
monkbroc: Any way to run the update code *after* the CLI command? That would create less disruption. It would also be possible to Ctrl-C the update check without issue.

lib/update-check.js
brycekahle: I think it would be very helpful when helping troubleshoot to have the version number there. It removes a round trip of us asking for the version. I did have it go to `stderr` instead so that if scripts were processing `stdout` they would not be affected.

app.js
brycekahle: If we reduce the interval to once per day, I think running beforehand is OK and makes sure the notice is seen.

lib/update-check.js
monkbroc: By reducing the roundtrip you mean when talking on the forum or support tickets? Hopefully the number of people using old versions will go down with this feature!

I still think the version number will be additional noise for regular users. How about putting the current version in `displayVersionBanner` instead?

brycekahle: Fair enough. What I really want is a log file when errors happen, which could contain the version number. This was just a stopgap until that happens.

PR #192, opened 2016-01-12
Electron key management support

brycekahle:
1. Adds `--protocol` argument for `keys` commands with valid values of `tcp` and `udp`. This argument controls DFU addresses, key algorithms, and more.
1. Adds `keys address` command that reads the server address, port, and protocol from the device.
1. Adds `keys protocol` command to set the transport protocol used when communicating with the cloud.
1. Removes Electron setup and directs the user to web setup.


joegoggins:
Yay! I'm excited for this to ship, I like the way interface landed here, good stuff!

PR #197, opened 2016-01-17
Use serialNumber to identify serial devices on Linux

### LINKED ISSUE #190: Electron detected as Core during setup on Ubuntu
The device specs don't match the Electron during setup on Ubuntu when it is blinking green or blinking blue (listening mode). It detects the Electron as a Core because [the manufacturer matches](https://github.com/spark/particle-cli/blob/bb3bf2204812e0d1948e618d78e5df49310d83e0/commands/SerialCommand.js#L111) but the device specs don't. I'll put more details here later.

monkbroc:
On (at least my) Linux Ubuntu 15.10, the reported vendor ID and product ID are the bus IDs instead of the device IDs. As a backup to the VID and PID identification, do a string matching on the `serialNumber` reported by the `serialport` module.

The expected strings for the Core, Photon and Electron are:
```
serialNumber: 'Spark_Devices_Spark_Core_with_WiFi_6D7238764957',
serialNumber: 'Particle_Photon_with_WiFi_00000000050C',
serialNumber: 'Particle_Electron_00000000050C',
```

Fixes #190

### DIFF
--- a/lib/deviceSpecs/specifications.js
+++ b/lib/deviceSpecs/specifications.js
@@ -2,7 +2,8 @@
 
 module.exports = {
 
-	'1d50:607f': { // Core
+	'1d50:607f': {
+		productName: 'Core',
 		tcpServerKey: {
 			address: '0x00001000',
 			size: 2048,
@@ -34,13 +35,14 @@ module.exports = {
 		},
 		serial: {
 			vid: '1d50',
-			pid: '607d'
+			pid: '607d',
+			serialNumber: 'Spark_Core'
 		},
-		productName: 'Core',
 		defaultProtocol: 'tcp',
 		productId: 0
 	},
-	'2b04:d006': { // Photon
+	'2b04:d006': {
+		productName: 'Photon',
 		tcpServerKey: {
 			address: '2082',
 			size: 512,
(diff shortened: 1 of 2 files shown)


brycekahle:
Thanks for this. Do you know if the hex number at the end varies at all?

monkbroc:
Apparently that number changes depending on the speed of the actual USB port:
https://github.com/spark/firmware/blob/develop/hal/src/photon/usbd_desc_device.h#L48

lib/deviceSpecs/specifications.js
```diff
@@ -34,13 +35,14 @@ module.exports = {
 		},
 		serial: {
 			vid: '1d50',
-			pid: '607d'
+			pid: '607d',
+			serialNumber: 'Spark_Devices_Spark_Core'
```
brycekahle: Based on https://github.com/spark/firmware/blob/develop/hal/src/core/usb_desc.c it looks like newer Core firmware might report as `Particle_Spark_Core_with_WiFi` so maybe we should reduce this to `Spark_Core`?

monkbroc: Good idea. I made the change.

PR #213, opened 2016-02-18
Warn user when variable/function name is too long

monkbroc:
See forum discussion http://community.particle.io/t/function-not-found/20189

I've done this often too: register a function/variable in code with a long name without checking the return value, then scratch my head for a while to figure out why I can't request its value.

This also adds a catch-all clause to monitorVariables to avoid this message when an error occurs:
```
Hit CTRL-C to stop!
Error: Variable not found
Potentially unhandled rejection [2] undefined (WARNING: non-Error used)
Error: Variable not found
Potentially unhandled rejection [4] undefined (WARNING: non-Error used)
Error: Variable not found
Potentially unhandled rejection [6] undefined (WARNING: non-Error used)
```

### DIFF
--- a/commands/VariableCommand.js
+++ b/commands/VariableCommand.js
@@ -52,6 +52,7 @@ VariableCommand.prototype = extend(BaseCommand.prototype, {
 	name: 'variable',
 	description: 'retrieve and monitor variables on your device',
 
+	maxNameLength: 12,
 
 	init: function () {
 		this.addOption('list', this.listVariables.bind(this), 'Show variables provided by your device(s)');
@@ -123,6 +124,11 @@ VariableCommand.prototype = extend(BaseCommand.prototype, {
 		var that = this;
 		if (!_.isArray(deviceId)) {
 			deviceId = [ deviceId ];
+		}
+
+		if(variableName.length > this.maxNameLength) {
+			console.log('Variable name is too long (maximum ' + this.maxNameLength + ' characters)');
+			return when.reject();
 		}
 
 		var api = new ApiClient();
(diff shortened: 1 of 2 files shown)


commands/VariableCommand.js
```diff
@@ -292,12 +303,16 @@ VariableCommand.prototype = extend(BaseCommand.prototype, {
...
-				});
+				}).catch(function() {});
 			}
 			checkVariable();
 		}).catch(function (err) {
 			self.error(err);
 		});
+	},
+
+	maximumVariableNameLength: function (deviceId) {
```
brycekahle: Unused?

commands/VariableCommand.js
```diff
@@ -272,6 +278,11 @@ VariableCommand.prototype = extend(BaseCommand.prototype, {
 		}
 		this.checkArguments(arguments);
 
+		if(variableName.length > this.maxNameLength) {
```
brycekahle: `deviceId` can also be a variable name at this point, so you should add another check inside of `disambiguateGetValue`.

commands/VariableCommand.js
monkbroc: Yeah, 1st implementation had maximum length as a function of deviceId but went back to a simple constant.

commands/VariableCommand.js
monkbroc: The error will be "no matching device" since the first argument is supposed to be a device.

I think that message is OK. The command line help is quite clear: "specify a device id (or all) and a variable name." I would never have thought that if a device is not specified it will pick one (which one?)

monkbroc:
After further thought and discussion on the forum, the best place for the name length check is as a firmware compile-time check so I'll close this change in the CLI.

PR #215, opened 2016-02-21
add: least heard for electron

kennethlimcp:
- shows in the locale GMT
- aligned with var/function tabspace

Example:

```
elec2g [xxx] (Electron)
  Last heard: Sun Feb 21 2016 08:41:36 GMT+0800 (SGT)
  Variables:
    soc (double)
    test (string)
  Functions:
    int digitalread(String args) 
    int digitalwrite(String args) 
    int analogread(String args) 
    int analogwrite(String args) 
elec3g [xx] (Electron)
  Last heard: Thu Feb 11 2016 10:00:25 GMT+0800 (SGT)
  Variables:
    test (string)
    soc (double)
  Functions:
    int digitalread(String args) 
    int digitalwrite(String args) 
    int analogread(String args) 
    int analogwrite(String args) 
oled [xx] (Photon) is online
  Functions:
    int stime(String args) 
    int reset(String args) 

````

### DIFF
--- a/commands/CloudCommands.js
+++ b/commands/CloudCommands.js
@@ -881,8 +881,15 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
 					name = chalk.cyan.dim(name);
 				}
 
-				var status = name + ' [' + device.id + ']' + deviceType + ' is ';
-				status += (device.connected) ? 'online' : 'offline';
+				var status = name + ' [' + device.id + ']' + deviceType;
+				if (deviceType === ' (Electron)') {
+					lines.push(status);
+					var lastHeard = new Date(device.last_heard).toLocaleString();
+					status = '  Last heard: ' + lastHeard;
+				} else {
+					status += ' is ';
+					status += (device.connected) ? 'online' : 'offline';
+				}
 				lines.push(status);
 
 				formatVariables(device.variables, lines);

### CODE REVIEW

**Comment by technobly** [comment]
This is great!  Mine show up as a bit different time format, and approx. 2 hours before my local time, but I'm technically GMT -06:00 so something is working differently for me.
```
electron11g [xxxx] (Electron)
  Last heard: 2/12/2016, 8:23:03 PM
electron6g [xxxx] (Electron)
  Last heard: 12/8/2015, 6:04:14 PM
photon7 [xxxx] (Photon) is offline
electron7 [xxxx] (Electron)
  Last heard: 2/21/2016, 9:08:12 PM
```

Would be nice if we could add something to adjust this time, even if we had to manually export a variable for the offset, it would be worth it.

kennethlimcp:
I wanted to add that to the json profile but thought the locale function might solve the headache. Seems like it might not work so well

brycekahle:
Thanks for the PR @kennethlimcp. Unfortunately `last_heard` from the API is extremely misleading. It does not update any time there is communication between the device and the cloud, because that would be a performance nightmare. In our database the field is called `last_handshake_at`, which is much more accurate. For TCP devices, this is somewhat useful. For UDP devices, even this time can be weeks out of date because we reuse session parameters to save data transfer.

Thus, we do not want to display `last_heard` prominently at this time. We are still discussing what is the best solution for the Electron. There was some miscommunication about implementation and where it stood, which is why it was incorrectly described in the Electron changelog.

technobly:
@kennethlimcp just tested this again and it seems to be correct with my local time.  Based on what Bryce is saying, that makes sense now... my Electron last night would have been off by two hours though due to when it last handshaked.  Just more data.

PR #216, opened 2016-02-21
device: add duo support

kennethlimcp:
- DFU works (tested sys1, sys2, user, factory)
- Serial works (wifi, inspect, identify)

- Not sure about the public/private keys address location though
- serialNumber: 'RedBear_Duo' is not verfied as this shows up in linux but should be correct based on https://github.com/redbear/firmware/blob/duo/bootloader/src/stm32f2xx/usbd_desc.c
- fixes https://github.com/spark/particle-cli/issues/173

### DIFF
--- a/lib/deviceSpecs/specifications.js
+++ b/lib/deviceSpecs/specifications.js
@@ -242,6 +242,51 @@ module.exports = {
 		},
 		defaultProtocol: 'udp',
 		productId: 10
+	},
+	'2b04:d058': {
+		productName: 'Duo',
+		tcpServerKey: {
+			address: '2082',
+			size: 512,
+			format: 'der',
+			alt: '1',
+			alg: 'rsa',
+			addressOffset: 384,
+			portOffset: 450
+		},
+		tcpPrivateKey: {
+			address: '34',
+			size: 612,
+			format: 'der',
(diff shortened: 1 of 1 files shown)


lib/deviceSpecs/specifications.js
```diff
@@ -242,6 +242,72 @@ module.exports = {
...
+		tcpServerKey: {
+			address: '2082',
+			size: 512,
+			format: 'der',
+			alt: '1',
+			alg: 'rsa',
+			addressOffset: 384,
+			portOffset: 450
+		},
+		udpServerKey: {
```
brycekahle: We should probably check, but I doubt they are supporting the UDP protocol.

dmiddlecamp: correct, as far as I know just TCP for the Duo

kennethlimcp: Can we just remove them? Will it cause any issues when running commands?

brycekahle: You can remove them, yes. The Core does not have them either.

kennethlimcp: @brycekahle, removed the UDP stuff.

PR #220, opened 2016-02-24
Added UDP listen command

wgbartley:
I chatted about this with @dmiddlecamp this morning (morning of 2016-02-24).  It's fleshing out a "stub" at https://github.com/spark/particle-cli/blob/master/commands/UdpCommands.js#L49 to give users an easy way to broadcast UDP messages from their devices and monitor them without resorting to complicated tools like tcpdump or Wireshark.


brycekahle:
Have you tried `nc -ul 5549`? That is pretty simple.

wgbartley:
@brycekahle - I have (it's what I used for testing before this PR), but utilities like netcat, tcpdump, and ngrep aren't something many Windows users will have installed (and know how to use).  This is a simpler, user-friendlier alternative.  ;-)

Issue #231, opened 2016-03-28
set max retry for wifi scan error


kennethlimcp:
When the photon is unable to perform a scan:

```
! PROTIP: non-broadcast network, please enter manual mode to proceed...

? Would you like to manually enter your Wi-Fi network configuration? No
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
> Your Photon encountered an error while trying to scan for nearby Wi-Fi networks. Retrying...
[...]

PR #236, opened 2016-04-19
Sets max limit for number of attempts to scan for Wi-Fi networks to 10

derekmpeterson:
Addresses issue #231


brycekahle:
I think we will want to transition to manual entry mode if this happens.

derekmpeterson:
@brycekahle Good call.  It will now transition back and ask if you'd like to use manual entry mode, or try scan mode again.

I also discovered my first commit introduced a bug that would cause the Wi-Fi scan to continue in the background even after informing the user that all 10 attempts had failed.  This has been fixed.

Issue #234, opened 2016-04-12
particle serial monitor "poll until serial port is available" feature


joegoggins:
I can capture serial logs using `particle serial monitor`, however, I miss the first part of the output because I need to time my execution of this command to be right after it opens the serial port, which is at least .5-1 seconds behind when the output starts, so I miss the first part.

Is there a good way to do this? Would adding a serial port poll flag like `--poll 1ms` to the `serial monitor` make sense??

The use case is something like this:

```
As a user debugging a complex Electron issue
I want to be able to capture the serial output logs from the very beginning and not miss the first second of serial output
```

m-mcgowan:
Implemented in the PR above, shipped in 1.19.4. Thanks @derekmpeterson!

PR #240, opened 2016-04-20
Implements the --follow option in the "serial monitor" command.

derekmpeterson:
Addresses #234 - Invoking the command with --follow will poll for an available serial device, and when one becomes available, immediately connect.  If the connection is interrupted, it will attempt to reconnect when the device is available again.

The monitor will reconnect automatically after rebooting or reflashing the device.  This makes debugging much more convenient.

While I was in there, in response to a TODO comment, I also added an interrupt handler to attempt to gracefully clean up the serial connection on SIGINT, SIGQUIT, SIGTERM, or exit.


monkbroc:
Thanks for the contribution. l need a little time to take a look at the code.

m-mcgowan:
Nice work @derekmpeterson, thanks!

Tested with and without the `--folllow` flag with 0, 1 and 2 devices connected to serial.

m-mcgowan:
This PR will automatically close once the changes are moved to master. Should have edited the base and changed from master to develop.

KravitzLab:
I have tried "particle serial monitor --follow" and after my photon goes to sleep and wakes up again, the command line doesn't monitor the serial COM anymore! This is the same problem with PuTTY as well...

kennethlimcp:
@KravitzLab something like CoolTerm should suit your needs. 

Which OS are you on? I will need to try this `--follow` flag on my side as well.

KravitzLab:
@kennethlimcp I got this solution off: https://community.particle.io/t/putty-or-other-serial-monitor-that-automatically-reconnects-windows/33579/7 check the comment made by rickkas7. I'll check CoolTerm out, but in case anyone else is reading this thread: Tera Term seems to be working as well. 

I am running Windows 7. Check it out and let me know what you get! I think this functionality is quite important to users developing low power systems, and debug Particle products after sleep.

PR #244, opened 2016-05-04
bump serialport dependency version for Node.js v6

nfriedly:
On Node.js v6, serialport was logging a huge warning for every particle-cli command, even ones that were unrelated to the serial port. See https://github.com/voodootikigod/node-serialport/issues/780 for details.

This change just bumps the serialport dependency to the latest release, which resolves the issue.

All the tests still pass and I did some basic manual testing and didn't see any issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
 - '0.12'
 - '4'
 - '5'
+- '6'
 sudo: false
 script:
   - npm run lint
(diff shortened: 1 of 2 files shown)


nfriedly:
Hum, looks like it's going to need a bit more work - now it's failing on older versions of node.js :(

nfriedly:
I also added node.js v6 to the test suite.

PR #248, opened 2016-05-12
Compile files in subdirectories

monkbroc:
- Update form-data module to pass the subdirectory information to the API
- Recursively traverse source directories
- Include the base name for all source files to be compiled


suda:
I think it looks great! We should rebase it on master and release :)
:shipit:

PR #250, opened 2016-05-24
integrates the cli-library-manager into the cli.

m-mcgowan:
Currently work in progress - basic listing of local and remote libraries and fetching remote libraries.

## Setup

- Ensure you have the `cli-library-manager` in the same directory as the `particle-cli` repo.
- Build should be running locally on port 3000, with the `feature/library_metadata` branch checked out or pulled

## Usage

see `./particle.sh help library`

All local library operations are performed relative to the current directory.

## Known issues

- Installing a library when it is already exists is currently not supported.


suda:
Ok, I tested it a bit and with https://github.com/spark/cli-library-manager/pull/2 merged in would be ready to :shipit: IMO.

Two additional notes:
* the bug I mentioned where it claimed the library doesn't exist was caused by cached `libs.json` endpoint which I fixed in https://github.com/spark/website/commit/f6029a172e289e3181ace6858c183ebbe9fb5f8d

PR #261, opened 2016-06-21
Add ES6 structure to CLI

monkbroc:
- Remove `particle.sh`: use `bin/particle.js` exclusively
- Put new ES6 files in `src` and they will be compiled to `dist`
- Entry point of CLI UI is `src/cli/app.js`
- Entry point of CLI used as a library is `src/lib/index.js`


m-mcgowan:
@monkbroc - this has already been merged during our work with libraries so closing this one.

PR #264, opened 2016-07-07
Add Oak and Bluz platforms

suda:


brycekahle:
Are the bluz gateway and beacon actual platforms in our system?

suda:
I believe so. They have separate `platform_id` specified.

PR #269, opened 2016-08-04
Improve list command

mebrunet:
- **Displays (Product product_id) when not of a known platform:** helpful for my team to know what product type the device is.
- **Applies list filters before requesting device attributes:** was noticing very long loading times when running `particle list <device_name>`. It seems the cli was nonetheless getting information from all devices.
- **Updates help info:** didn't even realize I could filter in the list command... very useful, I had been piping to grep before.


m-mcgowan:
fixed merge conflicts and merged to develop. Many thanks!

PR #278, opened 2016-09-03
Bugfix

joegoggins:
prevents calling `process.stdout._handle.setBlocking` when `_handle` is undefined (perhaps due to being run when there is no tty present?).

I was going to file an issue for this, but decided a PR would probably be easier for all parties :).


monkbroc:
Thanks. I merged this in. In case the CLI is used in scripts the output might be truncated due to node not waiting for all `console.log` output to be flushed before exiting.

PR #298, opened 2016-11-17
Disable update check through an environment variable

monkbroc:
Allow the CLI installer to start the CLI without the update check.

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -42,7 +42,7 @@ var settings = {
 	useSudoForDfu: false,
 	// TODO set to false once we give flags to control this
 	verboseOutput: true,
-	disableUpdateCheck: false,
+	disableUpdateCheck: envValueBoolean('PARTICLE_DISABLE_UPDATE', false),
 	updateCheckInterval: 24 * 60 * 60 * 1000, // 24 hours
 	updateCheckTimeout: 3000,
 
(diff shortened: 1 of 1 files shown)


settings.js
```diff
@@ -100,6 +100,22 @@ var settings = {
...
 };
 
+function envValue(varName, defaultValue) {
+	var value = process.env[varName];
+	return (typeof value === 'undefined') ? defaultValue : value;
+}
+
+function envValueBoolean(varName, defaultValue) {
+	var value = envValue(varName);
+	if (value === 'true' || value === 'TRUE' || value === '1') {
```
stefan-moraru: Are we sure we aren't going to have integer 1s?

settings.js
```diff
@@ -100,6 +100,22 @@ var settings = {
 	commandMappings: path.join(__dirname, 'mappings.json')
 };
 
+function envValue(varName, defaultValue) {
+	var value = process.env[varName];
+	return (typeof value === 'undefined') ? defaultValue : value;
+}
+
+function envValueBoolean(varName, defaultValue) {
+	var value = envValue(varName);
```
stefan-moraru: We should pass the defaultValue to 'envValue'

monkbroc: No, this is on purpose to get the raw value. The default handing is in `envValueBoolean`

settings.js
monkbroc: Environment variables are always strings.

PR #309, opened 2016-12-14
Compile library example

monkbroc:


monkbroc:
I'll change the acceptance test to use another app than Tinker. We already have too many versions of Tinker floating around and I'd like to avoid scattering some more.

PR #310, opened 2016-12-14
Feature/library ux

m-mcgowan:
consistent presentation of libraries between library list and library search.

also includes the library list commit since this wasn't previously PR'ed.


monkbroc:
Just needs some linting and fixing one test

PR #316, opened 2016-12-20
Feature/library list

m-mcgowan:
- allows the sections to list to be defined after the command `particle library list [sections...]`
- `--filter` parameter to filter the results by name
- pagination: after each page, hit enter for next, or CTRL+C to exit
- `--non-interactive` flag to turn off automatic pagination
- '--limit'  to specify the page size
- '--page' to specify the start page (1 based)

### DIFF
--- a/oldlib/prompts.js
+++ b/oldlib/prompts.js
@@ -97,7 +97,7 @@ var that = {
 		return dfd.promise;
 	},
 
-	enterToContinueControLCToExit(message) {
+	enterToContinueControlCToExit(message) {
 		if (!message) {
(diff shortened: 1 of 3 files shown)


oldlib/prompts.js
```diff
@@ -94,6 +97,24 @@ var that = {
 		return dfd.promise;
 	},
 
+	enterToContinueControLCToExit(message) {
```
monkbroc: There's a capital L too many here and in the reference below: enterToContinueControlCToExit

monkbroc:
I tried the branch and it worked well. Nice.

PR #319, opened 2016-12-22
library view command

m-mcgowan:


monkbroc:
I'm getting `Cannot read property 'then' of undefined` when I run `particle library view neopixel` I'm not sure why

Issue #280, opened 2016-09-17
Incorrect key pulled for Electron using TCP


towynlin:
Given I have an Electron using TCP
When I run `particle keys save electron-key-filename`
Then I expect to receive the RSA key for the TCP server
But instead I currently receive the EC key for the UDP server

brycekahle:
Community member just ran into this bug: https://community.particle.io/t/tls-key-for-local-server-cant-be-converted-to-der-format/27258

m-mcgowan:
The keys commands expects a `--protocol tcp` argument. So, not sure this is a bug other than missing documentation?

So the question is what to do when `--protocol` is not provided. Here's a few options:
- for devices with a single protocol, default to that (current behaviour on Core/Photon/P1)
- for devices with multiple protocol support:
 - raise an error if `--protocol` is not specified. This makes it unambiguous. 
 - or, inspect the device to determine the current protocol in use and default to that, raising the error above if for some reason the protocol cannot be determined.

PR #338, opened 2017-02-06
Feature/keys protocol

### LINKED ISSUE #279: Server key address output incorrect for Electron using TCP
Given I have an Electron with valid keys for the UDP and TCP servers
When I run `particle keys protocol tcp`
And I run `particle keys address`
Then I expect the output to be `tcp://device.spark.io:5683`
But instead I currently see `tcp://$id.udp.particle.io:5683`

It seems as though the [CLI implementation of `particle keys address`](https://github.com/spark/particle-cli/blob/b64252a42eae7b89b222011554a58777d29e8b08/commands/KeyCommands.js#L588-L595) is correctly checking the protocol and correctly changing the port, but incorrectly handling the host name, allowing the CLI default even to be overridden incorrectly.

m-mcgowan:
fixes #279 and #280 by reading the current protocol from the device when it is not specified, and the device supports multiple protocols.

Other changes:

- `particle keys protocol` displays the currently configured protocol for a device
- removed multiple protocol support from Photon/P1 since these devices only support TCP out the box.
- dfu.readBuffer/writeBuffer methods to read/write via buffers.  Hides the file handling needed when working with DFU, simplifying client code (and makes testing simpler.


commands/KeyCommands.js
```diff
@@ -243,9 +272,12 @@ KeyCommands.prototype = extend(BaseCommand.prototype, {
 				//make sure our device is online and in dfu mode
 				return dfu.findCompatibleDFU();
 			},
+			function () {
+				return self.validateDeviceProtocol();
+			},
 			//backup their existing key so they don't lock themselves out.
 			function() {
-				var alg = self._getPrivateKeyAlgorithm() || 'rsa';
+				var alg = self._getPrivateKeyAlgorithm() || defaultKeyAlgorithm;
```
monkbroc: `defaultKeyAlgorithm` is not defined anywhere.

monkbroc:
There's also the failing Travis :bomb:

PR #341, opened 2017-02-14
native module preemptive loading

### LINKED ISSUE #322: Detect when serialport module fails to load
The serialport module is natively compiled so when a different node is used to run the CLI than was used during installation the CLI will fail to run with "the module version expected is 48 and it got 47" or similar.

Detect this situation by preemptively requiring serialport and catching any errors. Display a message to the user to reinstall the CLI in this case.

m-mcgowan:
Loads native modules as part of startup to capture any loading issues. If they fail to load, the user is prompted to reinstall the CLI. fixes #322

### DIFF
--- a/src/app/app.js
+++ b/src/app/app.js
@@ -212,10 +212,32 @@ export class CLI {
 		return false;
 	}
 
+	loadNativeModules(modules) {
+		let errors = [];
+		for (let module of modules) {
+			try {
+				require(module);
+			} catch (err) {
+				errors.push(`Error loading module '${module}': ${err.message}`);
+			}
+		}
+		return errors;
+	}
+
 	run(args) {
 		settings.transitionSparkProfiles();
 		settings.whichProfile();
 		settings.loadOverrides();
+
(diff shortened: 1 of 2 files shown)


src/app/app.js
```diff
@@ -212,11 +212,33 @@ export class CLI {
...
 		settings.whichProfile();
 		settings.loadOverrides();
 
+		const nativeErrors = this.loadNativeModules(settings.nativeModules);
+		if (nativeErrors.length) {
+			const log = require('./log');
+			for (let error of nativeErrors) {
+				log.error(error);
+			}
+			log.fatal('Please reinstall the CLI again to ensure the module is installed to match the version of node.');
```
monkbroc: I'd say simply:
Reinstall the CLI by running npm install -g particle-cli

m-mcgowan: that will work for windows users as well after using the installer?

monkbroc: People with the installer are not likely to see this error. This also works on Windows.

PR #343, opened 2017-02-15
Feature/serial setup

m-mcgowan:
various setup improvements:

- all prompts default to `Y`
- if the user chooses not to setup a locally connected USB device, continue scanning for wifi devices rather than exit. 
- does not show “Obtained magical secure claim code” when obtaining the claim code fails
- clarification on how to enter manual mode for non-broadcast networks
- adds a default prompt selection Yes when confirming Wi-Fi credentials
- removes the `Configuration complete: You have just won the internet prompt` since setup is not complete yet.
- removed prompt to switch back to the original network, and just do it automatically. 
- `--manual` flag to force manual wifi setup on OSes that don't have the wifi manager support.
- `--yes` flag to allow all questions to be answered yes by default.
- fix spinner bug after 9 retries of 
- use latest version of softap-setup which fixes issues relating to sockets not timing out (causing the CLI to hand indefinitely.)

### DIFF
--- a/commands/CloudCommands.js
+++ b/commands/CloudCommands.js
@@ -748,11 +748,45 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
 		});
 	},
 
-	logout: function () {
+
+	doLogout: function(keep, password) {
+		var allDone = when.defer();
 		var api = new ApiClient();
+
+		pipeline([
+			function () {
+				if (!keep) {
+					return api.removeAccessToken(settings.username, password, settings.access_token);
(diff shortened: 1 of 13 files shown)


m-mcgowan:
I added a couple of changes.  Most importantly `particle setup wifi` now runs the full setup, but forces wifi, even when a device is connected via USB.  `particle serial wifi` can be used to obtain what used to be `particle setup wifi` (which I always felt was a misnomer because it doesn't run the full setup process, but now it does.)

PR #348, opened 2017-03-02
adds tests that validate the firmware binaries

m-mcgowan:


m-mcgowan:
@monkbroc  could you please push the correct binaries to this branch so we see the build passing.

PR #349, opened 2017-03-02
fixes compatibility with older node versions

dmiddlecamp:
submitted to fix compatibility with older node versions, specifically fixes the following:

```
Error loading command .../particle-cli/commands/KeyCommands.js SyntaxError: Unexpected token (
Error loading command .../particle-cli/commands/SerialCommand.js SyntaxError: In strict mode code, functions can only be declared at top level or immediately within another function.
```

### DIFF
--- a/commands/KeyCommands.js
+++ b/commands/KeyCommands.js
@@ -210,7 +210,7 @@ KeyCommands.prototype = extend(BaseCommand.prototype, {
 		return this._makeNewKey(filename);
 	},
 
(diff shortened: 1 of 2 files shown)


dmiddlecamp:
weird!  any issue with merging this pr?

monkbroc:
Thanks for the fix. It's merged. No issues but the weirdness of having several branches and doing complex merges...

PR #355, opened 2017-03-25
Add token commands to readme

kennethlimcp:
For whatever interesting reason, these commands never got documented.

Need to add them to the docs as well


monkbroc:
Thanks!

PR #367, opened 2017-05-14
update particle-cli on README.md

alan707:
In addition to installing and uninstalling, I'd thought it made sense to also show users how to **_update_** particle-cli.  There is a section on Particle Docs on how to do this, but I thought it might be convenient to see it on the Github repo which is where I was looking for it and couldn't find it.


monkbroc:
Thanks. This dropped out of the radar for some reason.

PR #379, opened 2017-08-07
Device doctor command to make device healthy

monkbroc:
The Device Doctor upgrade system firmware, flashes the default Tinker app, reset the device and server keys, clear the Wi-Fi settings for a device connected through the USB port.

Try it out with `particle device doctor`.


monkbroc:
@m-mcgowan Let's try to do a live review together this week.

PR #391, opened 2017-11-07
Fixed support for --yes when using cloud commands

mikeyyuen:
Added fix for --yes support
Updated shared command parsing to consume command items

https://github.com/spark/particle-cli/issues/382

### DIFF
--- a/commands/CloudCommands.js
+++ b/commands/CloudCommands.js
@@ -105,12 +105,15 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
 
 
 	checkArguments: function (args) {
+		args = Array.prototype.slice.call(args);
+
 		this.options = this.options || {};
 
 		if (!this.options.saveBinaryPath && (utilities.contains(args, '--saveTo'))) {
 			var idx = utilities.indexOf(args, '--saveTo');
 			if ((idx + 1) < args.length) {
 				this.options.saveBinaryPath = args[idx + 1];
+				args.splice(idx, 2);
 			} else {
 				console.log('Please specify a file path when using --saveTo');
 			}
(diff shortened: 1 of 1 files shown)


commands/CloudCommands.js
```diff
@@ -125,16 +128,33 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
...
+			}
 		}
 		if (!this.options.verbose) {
 			var idx = utilities.indexOf(args, '--verbose');
-			this.options.verbose = idx>=0;
+			if (idx >= 0) {
+				this.options.verbose = true;
+				args.splice(idx, 1);
+			} else {
+				this.options.verbose = true;
```
m-mcgowan: shouldn't this be `false`?

commands/CloudCommands.js
```diff
@@ -125,16 +128,33 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
...
+			}
 		}
 		if (!this.options.noconfirm) {
 			var idx = utilities.indexOf(args, '--yes');
-			this.options.noconfirm = idx>=0;
+			if (idx >= 0) {
+				this.options.noconfirm = true;
+				args.splice(idx, 1);
+			} else {
+				this.options.noconfirm = true;
```
m-mcgowan: and the same here, shouldn't this be `false`?

commands/CloudCommands.js
```diff
@@ -125,16 +128,33 @@ CloudCommand.prototype = extend(BaseCommand.prototype, {
 			// todo - move args parsing to a dedicated portion of the command interface (bound to I/O, and separate
 			// from the main command logic.)
 			var idx = utilities.indexOf(args, '--local');
-			this.options.local = idx>=0;
+			if (idx >= 0) {
```
m-mcgowan: it would be good to factor out this repeating pattern of looking for an argument, setting an option and removing the argument if found.

mikeyyuen:
The issue this fixes is the flash command won't accept --yes at all no matter which position on the command line its placed. e.g.

```
$ particle --version
1.16.0
$ particle flash --yes three-sim-card firmware.bin
I couldn't find that file: three-sim-card
Flash device failed
undefined
```

Where `three-sim-card` should be the device name.

commands/CloudCommands.js
mikeyyuen: Yeh, that makes sense, will take a look at making that change!

mikeyyuen:
Hi, this has been updated as requested, how does it look now?

PR #403, opened 2018-03-03
Remove old argument parser

monkbroc:
Move all the old commands to the new argument parser and rip out the old argument parser.


monkbroc:
Reviewed with Wojtek

Issue #378, opened 2017-07-10
Unhelpful warning for WPA enterprise setup


kennethlimcp:
```
? Have you reconnected to the internet? Yes
> Obtained magical secure claim code.

> Hey! We successfully connected to Photon-HHJX

> Now to configure our precious Photon-HHJX

! PROTIP: If you want to skip scanning, or your network is configured as a
! PROTIP: non-broadcast network, please choose No to the next prompt to enter manual mode.

? Shall I have the Photon scan for available Wi-Fi networks? Yes
? Please select the network to which your Photon should connect: SUTD_Student
? EAP Type PEAP/MSCHAPv2
(node:62637) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: SAP.eapTypeValue is not a function

KENMBP:~ kennethlimcp$ p --version
1.23.1
```

Photon system firmware: 0.6.1

PR #414, opened 2018-05-07
Fix incorrect usage of SoftAP

andrewmacheret:
Fixes https://github.com/particle-iot/particle-cli/issues/378 - this issue seems to be caused by a change in the dependency SoftAP - the function `eapTypeValue` was changed from a static function to a prototype function.


monkbroc:
Awesome! Thanks!

PR #421, opened 2018-06-05
fixed udev rules file reference

luukschipperheyn:
Fixed broken reference to udev file.


monkbroc:
Neat. Thanks! This will roll out with the next release.

PR #422, opened 2018-06-07
Improve error handling

monkbroc:
This is a massive PR that improves error handling by throwing `VError` that can be chained. So we can show the user things like: `Cannot claim device: host unreachable` instead of just `Cannot claim device`. It also tries to normalize the API responses to show a human-readable message. This PR does not address the fact that there are 3 API clients in the CLI. This will be done in a separate PR.


src/cmd/binary.js
```diff
@@ -25,47 +25,39 @@ License along with this program; if not, see <http://www.gnu.org/licenses/>.
...
-				return dfd.reject();
+	inspectBinary(binaryFile) {
+		return new Promise((fulfill, reject) => {
+			if (!fs.existsSync(binaryFile)) {
+				throw new VError(`Binary file not found ${binaryFile}`);
 			}
+			const parser = new Parser();
+			parser.parseFile(binaryFile, (fileInfo, err) => {
+				if (err) {
+					return reject(new VError(ensureError(err), `Could not parse ${binaryFile}`));
```
stefan-moraru: _trying to understand JS better_

On line 40, the error is only thrown. Is this on purpose? If so, is there a difference between reject and throw in this case?

src/cmd/flash.js
```diff
@@ -17,153 +16,132 @@ const systemModuleIndexToString = {
...
-		this.options = options;
-	}
-
-	flash() {
-		if (!this.options.params.device) {
+	flash(device, binary, files, { usb, serial, factory, force, target, yes }) {
+		if (!device && !binary) {
 			// if no device nor files are passed, show help
-			return when.reject();
+			// TODO: Replace by UsageError
```
stefan-moraru: TODO

src/cmd/flash.js
```diff
@@ -17,153 +16,132 @@ const systemModuleIndexToString = {
...
+		}).then(() => {
+			//only match against knownApp if file is not found
+			let stats;
+			try {
+				stats = fs.statSync(binary);
+			} catch (ex) {
+				// file does not exist
+				binary = dfu.checkKnownApp(binary);
+				if (binary === undefined) {
+					throw new Error('file does not exist and no known app found.');
```
stefan-moraru: `new VError`?

src/cmd/flash.js
```diff
@@ -17,153 +16,132 @@ const systemModuleIndexToString = {
...
+			}
 
-				return when.promise((resolve, reject) => {
-					const parser = new ModuleParser();
-					parser.parseFile(firmware, (info, err) => {
-						if (err) {
-							return reject(err);
-						}
+			if (!stats.isFile()){
+				throw new Error('You cannot flash a directory over USB');
```
stefan-moraru: `new VError`? Not sure if we should replace it, are there cases where we should use Error?

src/cmd/function.js
```diff
@@ -8,36 +9,32 @@ class FunctionCommand {
...
-							available.push('  int ' + name + '(String args) ');
-						}
-					}
-
-					let status = device.name + ' (' + device.id + ') has ' + available.length + ' functions ';
-					if (available.length === 0) {
-						status += ' (or is offline) ';
+					for (let idx = 0; idx < device.functions.length; idx++) {
+						const name = device.functions[idx];
+						available.push('  int ' + name + '(String args) ');
```
stefan-moraru: Fancier with `

`  int  ${name}(String args) `

src/cmd/function.js
```diff
@@ -8,36 +9,32 @@ class FunctionCommand {
...
 					}
+				}
 
-					lines.push(status);
-					lines = lines.concat(available);
+				let status = device.name + ' (' + device.id + ') has ' + available.length + ' functions ';
```
stefan-moraru: `${device.name} (${device.id}) has ${available.length} functions `

src/cmd/subscribe.js
```diff
@@ -28,7 +22,7 @@ class SubscribeCommand {
 			eventLabel = 'all events';
 		}
```
stefan-moraru: Lines 10 -> 29

```
		let eventName = event[0];
		let deviceId = event[1] || device;
 		if (eventName === 'mine') {
 			eventName = null;
 		}
 
 		let eventLabel = eventName;
 		if (eventLabel) {
 			eventLabel = '"' + eventLabel + '"';
 		} else {
 			eventLabel = 'all events';
 		}
```

```
const eventName = event[0] === 'mine' ? null : event[0];
const eventLabel = eventName ? `"${eventName}"` : 'all events';
const deviceId = event[1] || device;
```

src/cmd/variable.js
```diff
@@ -87,86 +81,81 @@ class VariableCommand {
...
+			} else {
+				const promises = [];
+				for (let i = 0; i < devices.length; i++) {
+					const deviceid = devices[i].id;
+					if (devices[i].connected) {
+						promises.push(api.getAttributes(deviceid));
+					} else {
+						promises.push(Promise.resolve(devices[i]));
 					}
+				}
```
stefan-moraru: Lines 138 -> 146

```
				const promises = [];
				for (let i = 0; i < devices.length; i++) {
					const deviceid = devices[i].id;
					if (devices[i].connected) {
						promises.push(api.getAttributes(deviceid));
					} else {
						promises.push(Promise.resolve(devices[i]));
					}
				}
```

const promises = devices
.map(device => {
  return device.connected ? api.getAttributes(device.id) : Promise.resolve(device);
});

src/lib/ApiClient.js
```diff
@@ -991,6 +1008,34 @@ class ApiClient {
...
+			reason = response.errors.map((err) => {
+				if (err.error) {
+					if (err.error.status) {
+						return err.error.status;
+					} else {
+						return err.error;
+					}
+				} else {
+					return err;
+				}
```
stefan-moraru: Lines 1022 -> 1030.

We can replace the nested if statements with
```
return err.error && err.error.status || err.error || err; 
```
image

src/cmd/binary.js
monkbroc: No, line 40 is a bug.

monkbroc: In general I prefer to throw errors in both async and sync code so I don't have to use 2 patterns for what feels like the same thing. The challenge comes in when mixing callback-based code with Promise based code. It can be easy to make a mistake.

src/cmd/flash.js
monkbroc: This works, but it's a quirk of the implementation that I'd like to make explicit: if a command rejects without a reason (like `return Promise.reject()`) the command processor will show usage information. We should have a shared UsageError class that can be thrown to accomplish this.

src/cmd/flash.js
monkbroc: In this case there's no difference. VError is a class that add additional functionality on top of error, but if you don't use that functionality, using Error is OK. Node error handling is quite confusing.

src/cmd/function.js
monkbroc: Yeah, most of this code predates the backtick operator and I didn't refactor all the instances of string concatenation. It's work left for a future maintainer :)

src/cmd/variable.js
monkbroc: I don't think you even need the `Promise.resolve`, just return `device` in the second case.

src/lib/ApiClient.js
monkbroc: I'd like to see some parentheses since the priority of operators tends to play tricks with these long boolean chains.

monkbroc:
Stefan, thanks for the review! Go ahead and push a commit with your suggested changes.

src/lib/ApiClient.js
stefan-moraru: `return (err.error && err.error.status) || err.error || err; `

monkbroc:
We created this story as part of the review https://app.clubhouse.io/particle/story/16307/add-tests-for-cloud-compile-file-mapping

monkbroc:
Stefan, I created a story for you to implement the improvements you suggested https://app.clubhouse.io/particle/story/16329/implement-review-comments-from-422

PR #429, opened 2018-06-22
bugfix/project-create-args

busticated:
https://app.clubhouse.io/particle/story/16775/dir-parameter-and-name-option-are-ignored-in-project-create-command


src/cli/project_init.js
```diff
@@ -155,9 +155,11 @@ class CLIProjectInitCommandSite extends ProjectInitCommandSite {
 }
 
 export function command(argv) {
-	const dir = argv.params.dir || process.cwd();
-	const site = new CLIProjectInitCommandSite(dir);
 	const cmd = new ProjectInitCommand();
+	const site = new CLIProjectInitCommandSite({
+		directory: argv.params.dir,
+		name: argv.name
+	});
```
busticated: some context: arg handling has been broken for a while it seems (since https://github.com/particle-iot/particle-cli/commit/9ea72b039310ebabee00e94b6519cf1366cbb5d5#diff-dc2a887cded43f37a650c27b9b32b44dR35). dropping `process.cwd()` here has no end-user impact since it was redundant - we call it again here: https://github.com/particle-iot/particle-cli/blob/efac328380f937f2eef001c788d68f5e15972b6f/src/cli/project_init.js#L83

so overall this seems pretty low-risk 👍

PR #431, opened 2018-06-25
Fix issue with function and variable commands

monkbroc:
* Fix server error when Particle.function returns 0
* Only print timestamp when using the --time flag


test/cmd/function.spec.js
```diff
@@ -0,0 +1,109 @@
...
+		return stubs.api;
+	}
+};
+
+const FunctionCommand = proxyquire('../../src/cmd/function', {
+	'../lib/ApiClient.js': stubs.ApiClient
+});
+
+
+describe.only('Function Command', () => {
```
busticated: remove `only`

PR #444, opened 2018-08-10
Add Xenon support, set initial response timeout to 10 seconds

YutingYou:
### Feature

Support Xenon YModem firmware updating, set initial response timeout to 10 seconds because it takes about 6 seconds to send initial response on Xenon.


monkbroc:
Press the Merge pull request button if this is ready.

PR #450, opened 2018-09-08
Add command to preprocess ino file

monkbroc:
Allows to preprocess ino file into cpp file easily outside of the cloud compiler.

Includes tests for the argument parsing, implementation and an acceptance test


monkbroc:
Reviewed with Anatoly

PR #464, opened 2018-12-18
bugfix/login-token-spinner

busticated:
https://app.clubhouse.io/particle/story/24297/cli-login-using-token-not-working


busticated:
> Want to bump the patch version and release this right away?

yar 👍

PR #466, opened 2018-12-20
Support NCP firmware in DFU flash

monkbroc:
Story details: https://app.clubhouse.io/particle/story/26678

Add support for flashing NCP firmware through DFU.

Steps to test:
```
npm install
bin/particle.js flash --usb argon-ncp-firmware-0.0.5-ota.bin
```


monkbroc:
Ok. I took the command from an old message from you on Slack but I guess this doesn't work anymore. Can you think of a strategy for updating Gen 3 device firmware with the CLI? Consider the bootloader, tinker and NCP. Currently the `particle update` command is only able to do DFU flash. It might be OK to avoid updating the NCP and just provide additional instructions for the user to run `particle flash --serial` to update the NCP

avtolstoy:
I'm not really sure where you got it, we've never supported updating the NCP that way, since it has its own flash which is not directly exposed in DFU 😅 

I'd vote for an additional step out of DFU with `--serial`, otherwise we'll need to add some additional logic in the bootloader to notify DeviceOS that there is a pending NCP image waiting in the OTA area and would have to trigger an update on system firmware boot. We can technically do this, but the question is do we want to spend time implementing this, considering the fact that devices runnng anything under the release supporting this will anyway require several steps (manually updating the bootloader beforehand, or at least performing an additional reset to update it).

monkbroc:
Closing as it's not currently possible to flash NCP through DFU

PR #470, opened 2019-01-24
particle keys server uses wrong host address for Gen 3 devices

monkbroc:
Story details: https://app.clubhouse.io/particle/story/27508


src/lib/deviceSpecs/specifications.js
```diff
@@ -383,6 +383,59 @@ const specs = {
...
+		},
+		systemFirmwareOne: {
+			address: '0x00030000',
+			alt: '0'
+		},
+		knownApps: {
+		},
+		serial: {
+			vid: '2b04',
+			pid: 'c017',
```
anatoly-particle: Shouldn't it be `c016 ` here?

monkbroc: Good catch!

PR #471, opened 2019-02-08
added --yes for particle flash serial to remove blue light prompt

markterrill:
Hi,

Not being able to do an unintended install/upgrade to 1.0.0 is a PITA. This makes the bootloader update via particle flash serial work fine.

Previously signed CLA in '16


monkbroc:
Great call. Thanks for adding this.

PR #473, opened 2019-02-11
Mesh commands

sergeuz:
This PR implements the following new commands:

#### Mesh network management

Create a new network:
```
particle mesh create <network name> <device> [--channel=N] [--password=...] [--yes]
```
Add a device to the current network of an assisting device:
```
particle mesh add <new device> <assisting device> [--password=...] [--yes]
```
Remove a device from its current network:
```
particle mesh remove <device> [--yes]
```
List all registered networks and their devices:
```
particle mesh list [network] [--networks-only]
```
Get the current device's network:
```
particle mesh info <device>
```
Scan for networks:
```
particle mesh scan <device>
```

#### Misc. USB commands

List Particle USB devices attached to the host:
```
particle usb list [--exclude-dfu] [--ids-only]
```
Put a device or multiple devices into the listening mode:
```
particle usb start-listening [devices...] [--all]
```
Make a device or multiple devices exit the listening mode:
```
particle usb stop-listening [devices...] [--all]
```
Put a device or multiple devices into the safe mode:
```
particle usb safe-mode [devices...] [--all]
```
Put a device or multiple devices into the DFU mode:
```
particle usb dfu [devices...] [--all]
```
Reset a device or multiple devices:
```
particle usb reset [devices...] [--all]
```
Install udev rules for Particle USB devices (Linux-only):
```
particle usb configure
```

Note: This PR is a WIP and there's a number of known bugs, limitations and TODOs:

- ~The `IN control transfer failed` error is back and needs to be debugged.~
- Gateway devices are expected to be configured via the mobile apps (initial device setup is out of the scope of this PR).
- ~Manual/automatic udev configuration is not implemented yet (`particle usb configure`?).~
- ~This branch depends on the feature branches of [`particle-usb`](https://github.com/particle-iot/particle-usb) and [`particle-api-js`](https://github.com/particle-iot/particle-api-js/tree/feature/mesh_api) which haven't been merged or published yet.~
- Unit tests are missing.


sergeuz:
[ch27463]

src/cmd/mesh.js
```diff
@@ -0,0 +1,340 @@
...
+	return /^[0-9a-f]{24}$/i.test(name);
+}
+
+export class MeshCommand {
+	constructor(settings) {
+		this._apiToken = settings.access_token;
+		this._api = new ParticleApi(settings.apiUrl, { accessToken: this._apiToken }).api;
+	}
+
+	async create(args) {
```
monkbroc: We don't use `async/await` yet in the CLI since we still support Node < 10. I know we could add support by changing the Babel configuration but when we did that we had to ship the CLI with the Babel polyfill or the Babel runtime and we ran into issues with the polyfill being required multiple times in certain situations. When we last considered this we decided to stick to vanilla Promises for the time being in the CLI.

src/cli/mesh.js
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return meshCommand._instance;
+}
+
+export default ({ commandProcessor, root }) => {
+	const mesh = commandProcessor.createCategory(root, 'mesh', 'Manage mesh networks');
+
+	commandProcessor.createCommand(mesh, 'create', 'Create a new network', {
+		// TODO: Provide descriptions for positional arguments?
+		params: '<network name> <device>',
```
monkbroc: Let's use snake_case for the parameters so they can be referenced using `params.network_name` in later code.

src/cmd/mesh.js
sergeuz: Do you remember what was the issue with babel-runtime specifically? The CLI, including mesh commands, seems to work fine with Node 6 and the following babel config:
```
{
  "presets": ["env"],
  "plugins": ["transform-runtime", "add-module-exports"]
}
```
I'm fine with rewriting this code, I just find `await` especially helpful in tools like the CLI, where all the operations are typically done sequentially

monkbroc: I don't recall the problem specifically. It was something with double includes of babel modules due to the different babel config in particle-api-js, particle-library-manager and particle-cli (if I recall correctly) crashing the CLI. Removing the need to load babel-polyfill or babel-runtime got rid of the issues. Some of these issues were reported by customers since they happened only in certain cases that we weren't testing.

I'm also a bit wary of adding a new coding style to an existing application without a plan to migrate the rest of the code base to fit the new style.

If we add async/await to the CLI, I think it should be done with `babel-plugin-transform-async-to-generator` since it brings a lot less bagage than `babel-plugin-transform-runtime` (and we don't need most of the stuff that transform-runtime gives). For example, `transform-runtime` converts `Object.assign` to a babel function which is unnecessary.

busticated: > I'm also a bit wary of adding a new coding style to an existing application without a plan to migrate the rest of the code base to fit the new style.

very much agree.

> If we add async/await to the CLI, I think it should be done with `babel-plugin-transform-async-to-generator`

+1 here as well but also in general transpiling is kind of the worst so the less we do overall, the better imo.

that all said, Node v6 LTS hits end of life "April 2019" ([schedule](https://github.com/nodejs/Release)) and Node v8 LTS supports `async` / `await` natively ([docs](https://node.green/#ES2017-features-async-functions-await)) so fwiw

sergeuz: > If we add async/await to the CLI, I think it should be done with `babel-plugin-transform-async-to-generator`

> +1 here as well but also in general transpiling is kind of the worst so the less we do overall, the better imo

If we do care about transpiling and extra stuff being pulled into the CLI, shouldn't we simply switch to `preset-env` and rely on babel to use only those transformations and polyfills which are required for the target Node version?

> I'm also a bit wary of adding a new coding style to an existing application without a plan to migrate the rest of the code base to fit the new style

`async/await` aside the CLI is already written in different styles. It's not that I want to make things worse, I just don't think that starting to use `async/await` somewhere forces us to rethink the design of the existing code. In my opinion, it's fine to use plain promises and async functions in the same codebase, moreover, you don't actually have an alternative to promises if you want to do things concurrently.

Anyway, by no means I want to start arguing over established practices in the codebase with which I'm new to. Let me know if I should rewrite the code to get this PR merged, I'm totally fine with that

busticated: > shouldn't we simply switch to `preset-env` and rely on babel to use only those transformations and polyfills

as mentioned [above](https://github.com/particle-iot/particle-cli/pull/473#discussion_r255965192), polyfills have been problematic in the past. introducing `preset-env` as part of a larger, separate clean-up effort would probably be fine but then again dropping support for < node 8 and eliminating all pre-processing / transpiling would be even better imo.

that's all a bit of a digression though. as i would expect you'd agree, your effort here should be as focused as possible - overhauling the build system seems like it should be out of scope if only for your own sanity.

sergeuz: > that's all a bit of a digression though. as i would expect you'd agree, your effort here should be as focused as possible - overhauling the build system seems like it should be out of scope

That's definitely not my intent, and I'd be happy to make a practical decision based on this discussion too. Given that polyfills are not desirable in the CLI, we have two options:
- Enable the `async-to-generator` plugin as the least invasive way to get support for `async/await` in the CLI.
- Rewrite the code to avoid using `async/await`.

I doubt the second option will take more than an hour to implement, just let me know if that's a preferred way forward

busticated: > let me know if that's a preferred way forward

i defer to @monkbroc but my gut tells me that after wrestling w/ `npm i`, `package-lock.json` and all the additional testing that comes along with it, rewriting to avoid `async/await` will prove to be the simplest / least risky route.

fwiw, i'm also fine with enabling the `async-to-generator` plugin but there be dragons 🐲🐉 

either way, thanks for wading through all of this annoying stuff. i would like very much to get things into a better state soon 👍

src/cli/usb.js
```diff
@@ -0,0 +1,109 @@
...
+	});
+
+	// Common options for start-listening, stop-listening, safe-mode, dfu and reset
+	const commonOptions = {
+		'all': {
+			description: 'Send the command to all devices connected to the host computer',
+			boolean: true
+		},
+		'one': {
+			description: 'Send the command to a single device connected to the host computer',
```
monkbroc: In other CLI commands, this mode is the default. If you do `particle flash --usb` it just picks the one device connected. I'd like to avoid the need to specify a device ID if there is only one device connected.

src/cli/usb.js
```diff
@@ -0,0 +1,109 @@
...
+		examples: {
+			'$0 $command my_device': 'Reset a device named "my_device"',
+			'$0 $command --all': 'Reset all devices connected to the host computer'
+		},
+		handler: (args) => {
+			return usbCommand().reset(args);
+		}
+	});
+
+	commandProcessor.createCommand(usb, 'configure', 'Update the system USB configuration', {
```
monkbroc: From the user's perspective, what does this do? What needs to be configured?

sergeuz: This command installs the udev rules on Linux. @busticated needed it for Workbench

monkbroc: Ah, ok. Let's merge it as is, but I'd like to add a bit more description later about what this does.

monkbroc:
I ran into an issue installing the particle-usb module on Windows through the CLI wrapper. Prebuild used the system version of Node instead of the wrapper so it installed the wrong dependency.

<details>
 <summary>Install Log</summary>

```
C:\Users\monkbroc\AppData\Local\particle>node-v8.15.0-windows-x64\bin\node.exe node-v8.15.0-windows-x64\lib\node_modules\npm\bin\npm-cli.js install h:\Downloads\particle-cli-1.38.0.tgz
npm WARN lifecycle The node binary used for scripts is C:\Program Files\nodejs\node.exe but npm is using C:\Users\monkbroc\AppData\Local\particle\node-v8.15.0-windows-x64\bin\node.exe itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.

> usb@1.5.0 install C:\Users\monkbroc\AppData\Local\particle\node_modules\usb
> prebuild-install --tag-prefix particle- --verbose || node-gyp rebuild

prebuild-install info begin Prebuild-install version 5.2.5
prebuild-install info looking for cached prebuild @ C:\Users\monkbroc\AppData\Roaming\npm-cache\_prebuilds\187816-usb-v1.5.0-node-v64-win32-x64.tar.gz
prebuild-install http request GET https://github.com/particle-iot/node-usb/releases/download/particle-1.5.0/usb-v1.5.0-node-v64-win32-x64.tar.gz
prebuild-install http 200 https://github.com/particle-iot/node-usb/releases/download/particle-1.5.0/usb-v1.5.0-node-v64-win32-x64.tar.gz
prebuild-install info downloading to @ C:\Users\monkbroc\AppData\Roaming\npm-cache\_prebuilds\187816-usb-v1.5.0-node-v64-win32-x64.tar.gz.9396-356bb31c99474.tmp
prebuild-install info renaming to @ C:\Users\monkbroc\AppData\Roaming\npm-cache\_prebuilds\187816-usb-v1.5.0-node-v64-win32-x64.tar.gz
prebuild-install info unpacking @ C:\Users\monkbroc\AppData\Roaming\npm-cache\_prebuilds\187816-usb-v1.5.0-node-v64-win32-x64.tar.gz
prebuild-install info unpack resolved to C:\Users\monkbroc\AppData\Local\particle\node_modules\usb\build\Release\usb_bindings.node
prebuild-install info unpack required C:\Users\monkbroc\AppData\Local\particle\node_modules\usb\build\Release\usb_bindings.node successfully
prebuild-install info install Successfully installed prebuilt binary!

> protobufjs@6.8.8 postinstall C:\Users\monkbroc\AppData\Local\particle\node_modules\protobufjs
> node scripts/postinstall

npm WARN saveError ENOENT: no such file or directory, open 'C:\Users\monkbroc\AppData\Local\particle\package.json'
npm WARN enoent ENOENT: no such file or directory, open 'C:\Users\monkbroc\AppData\Local\particle\package.json'
npm WARN particle No description
npm WARN particle No repository field.
npm WARN particle No README data
npm WARN particle No license field.

+ particle-cli@1.38.0
added 33 packages from 79 contributors, updated 1 package and audited 9050 packages in 11.525s
found 36 vulnerabilities (15 low, 21 moderate)
  run `npm audit fix` to fix them, or `npm audit` for details

C:\Users\monkbroc\AppData\Local\particle>particle version
1.38.0

C:\Users\monkbroc\AppData\Local\particle>particle usb
Control USB devices
Usage: particle usb <command>
Help:  particle help usb <command>

Commands:
  list             List the devices connected to the host computer
  start-listening  Put a device into the listening mode
  stop-listening   Make a device exit the listening mode
  safe-mode        Put a device into the safe mode
  dfu              Put a device into the DFU mode
  reset            Reset a device
  configure        Update the system USB configuration

Global Options:
  -v, --verbose  Increases how much logging to display                                                           [count]  -q, --quiet    Decreases how much logging to display                                                           [count]

C:\Users\monkbroc\AppData\Local\particle>particle usb list
! Please reinstall the CLI again using npm install -g particle-cli
The module '\\?\C:\Users\monkbroc\AppData\Local\particle\node_modules\usb\build\Release\usb_bindings.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 64. This version of Node.js requires
NODE_MODULE_VERSION 57. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```

</details>

This can be fixed by adding `NPM_CONFIG_SCRIPTS_PREPEND_NODE_PATH=true` to https://github.com/particle-iot/particle-cli-wrapper/blob/364995b6cf8d4b67fc680f86915226e2fd72fa19/gode/packages.go#L142 and releasing a new version of the CLI wrapper.

I also ran into the Windows driver issue on Windows 10. Let's discuss the way forward with @avtolstoy 

With the libwdi driver installed by the CLI installer:
```
C:\Users\monkbroc\AppData\Local\particle>particle usb list
Unable to open USB device: LIBUSB_ERROR_NOT_SUPPORTED
```

After uninstalling the libwdi driver (but without needing to use Zadig):
```
C:\Users\monkbroc\AppData\Local\particle>particle usb list
<no name> [111111111111111111111111] (Photon)
```

monkbroc:
Scratch that piece about changing the CLI wrapper. I was trying to manually install the package and my method was not representative. The install will work fine as is once the new version of the CLI module is published.

PR #481, opened 2019-04-16
Runtime Device OS logging configuration

sergeuz:
**Note: The feature implemented in this PR relies on a new Device OS functionality that has not been released yet: [device-os/#1748](https://github.com/particle-iot/device-os/pull/1748).**

This PR implements a new command called `log` that allows configuring Device OS logging at runtime. The detailed description of the command can be found in the [PRD](https://docs.google.com/document/d/14KplDrRw7Q0WNuHYpfyxIxKee_9QICiPezUa9x1Hqcc/edit?usp=sharing).

Usage:

- Enabling logging on the default `Serial` interface:
```
particle log my_device
```

- Enabling logging on the `Serial1` interface:
```
particle log my_device Serial1
```

- Showing only application messages and system errors:
```
particle log my_device -l error -f app:all
```

Example output:

pretty

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -24,6 +24,7 @@ const webhook = require('./webhook');
 const whoami = require('./whoami');
 const mesh = require('./mesh');
 const usb = require('./usb');
+const log = require('./log');
 
 /**
  * The default function export from this module registers all the available commands.
@@ -66,5 +67,6 @@ module.exports = function registerAllCommands(context) {
 	whoami(context);
 	mesh(context);
 	usb(context);
+	log(context);
 	alias(context);
 };
--- a/src/cli/log.js
+++ b/src/cli/log.js
@@ -1,0 +1,39 @@
+const settings = require('../../settings');
+
+function logCommand() {
+	if (!logCommand._instance) {
+		const LogCommand = require('../cmd/log');
+		logCommand._instance = new LogCommand(settings);
+	}
+	return logCommand._instance;
+}
+
+module.exports = ({ commandProcessor, root }) => {
+	commandProcessor.createCommand(root, 'log', 'Display log messages from a device', {
+		params: '<device> [stream] [serial_port]',
+		options: {
+			'level': {
+				description: 'Default logging level',
+				string: true,
+				alias: 'l'
+			},
+			'filter': {
+				description: 'Category filter',
+				string: true,
+				alias: 'f'
+			},
+			'baud': {
+				description: 'Baud rate',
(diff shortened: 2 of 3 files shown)


sergeuz:
[ch31041]

sergeuz:
[ch30129]

monkbroc:
From a product perspective here's what I see when I test:

When pressing reset or pulling the USB port I get a red message saying `bad file descriptor`. It would be nicer to see "connection closed" or something.

I cannot get the feature to work on a Photon, when compiling Device OS from https://github.com/particle-iot/device-os/pull/1748

monkbroc:
I can't figure out how to use `Serial1`. It just says `Serial port is not found`.

monkbroc:
Ah, for the serial not found I see the problem. I use a Photon with this serial echo program as my UART -> USB converter:
```
SYSTEM_MODE(MANUAL);

void setup() {
  Serial1.begin(115200);
}

void loop() {
  while(Serial1.available()) {
    char c = Serial1.read();
    Serial.write(c);
  }
}
```

Maybe just adding another example for Serial1 to show how to select a custom USB port in case the auto-detection doesn't find the right one would be enough?

monkbroc:
I'm also not able to get any logs from `USBSerial1`. When I do 
```
$ particle log my_device USBSerial1
Opening serial port: /dev/ttyACM0
```
I get no logs.

It opens the same serial port as doing 
```
$ particle log my_device Serial
Opening serial port: /dev/ttyACM0
```

busticated:
@sergeuz i rebased this branch against latest master. if you have work in-fligth, be sure to `git checkout -b feature/log_config_bkup` _before_ pulling.

sergeuz:
@monkbroc I fixed the code according to your suggestions and rebased this branch. I had to disable `USBSerial1` for now, as it turned out that calling `USBSerial1.begin()` in the Device OS causes the device to be detached from the system (or otherwise invalidates the libusb handle opened by particle-usb), which in turn causes the logging request to fail

sergeuz:
> I cannot get the feature to work on a Photon

I accidentally enabled the new logging requests only on mesh platforms. This is now fixed in https://github.com/particle-iot/device-os/pull/1748

busticated:
@sergeuz what's the status of this work? are we still waiting for Device OS side updates?

busticated:
this appears to be orphaned - closing. feel free to re-open when it's actionable 👍

WakfuYT:
.

PR #485, opened 2019-05-06
Mesh command fixes

sergeuz:
This PR introduces the following changes:

- When adding an unclaimed device to a network, make sure the device is removed from its current network first (see [ch30101](https://app.clubhouse.io/particle/story/30101/half-initialized-mesh-networks-make-mesh-device-unable-to-be-provisioned) for the details).
- Filter out duplicate entries from the list of discovered networks (see [this](https://s.slack.com/archives/CG6D5KB43/p1555091439015800?thread_ts=1554999936.004900&cid=CG6D5KB43) discussion and the related [fixes](https://github.com/particle-iot/device-os/pull/1760) in the Device OS).
- Remove the workaround for the API service issue that was fixed in https://github.com/particle-iot/api-service/pull/601.


sergeuz:
[ch31180]

PR #493, opened 2019-06-11
adds support for radio stack modules

avtolstoy:
### Description

This PR:
1. adds support for 'radio stack' modules
2. adds support for `DROP_MODULE_INFO` module flag introduced in https://github.com/particle-iot/device-os/pull/1816 when flashing over DFU
3. prints secondary module dependency information in `binary inspect`

### Testing

Follow the guide in https://github.com/particle-iot/device-os/pull/1816, or use the following pre-built [Argon binaries](https://github.com/particle-iot/particle-cli/files/3276558/argon-binaries.zip) and flash `system-part1` and `bootloader` to the device.

1. `particle binary inspect argon-softdevice-6.1.1.bin` should work correctly:
```console
 CRC is ok (78715fba)
 Compiled for argon
 This is a radio stack module number 0 at version 182
 It depends on a system module number 1 at version 1300
 It also depends on a bootloader number 0 at version 311
```
2. `particle serial inspect` should work correcty and should report a radio stack module with version 169 (SoftDevice 6.0.0)
```console
Platform: 12 - Argon
Modules
  Bootloader module #0 - version 311, main location, 49152 bytes max size
    Integrity: PASS
    Address Range: PASS
    Platform: PASS
    Dependencies: PASS
  System module #1 - version 1300, main location, 671744 bytes max size
    Integrity: PASS
    Address Range: PASS
    Platform: PASS
    Dependencies: PASS
      Bootloader module #0 - version 311
  User module #1 - version 6, main location, 131072 bytes max size
    UUID: A7979068CD4812DA30511222456D3E613D8F5AA38DB4D780000565E85F2ECF97
    Integrity: PASS
    Address Range: PASS
    Platform: PASS
    Dependencies: PASS
      System module #1 - version 1300
  NCP module #0 - version 5, main location, 1536000 bytes max size
    Integrity: PASS
    Address Range: PASS
    Platform: PASS
    Dependencies: PASS
  Radio stack module #0 - version 169, main location, 196608 bytes max size
    Integrity: PASS
    Address Range: PASS
    Platform: PASS
    Dependencies: PASS
```
3. `particle flash --usb argon-softdevice-6.1.1.bin` should update the SoftDevice to 6.1.1 and it should boot normally. Running `particle serial inspect` afterwards should report version 182 (SoftDevice 6.1.1) for radio stack module.

### References
- [CH31496]
- [DeviceOS PR](https://github.com/particle-iot/device-os/pull/1816)
- [binary-version-reader PR](particle-iot/binary-version-reader#24)

### DIFF
--- a/src/cmd/flash.js
+++ b/src/cmd/flash.js
@@ -2,12 +2,11 @@ const fs = require('fs');
 const VError = require('verror');
 const dfu = require('../lib/dfu');
 const ModuleParser = require('binary-version-reader').HalModuleParser;
+const ModuleInfo = require('binary-version-reader').ModuleInfo;
 const deviceSpecs = require('../lib/deviceSpecs');
 const ensureError = require('../lib/utilities').ensureError;
-
-const MONOLITHIC = 3;
-const SYSTEM_MODULE = 4;
-const APPLICATION_MODULE = 5;
+const temp = require('temp').track();
+const when = require('when');
 
 const systemModuleIndexToString = {
 	1: 'systemFirmwareOne',
@@ -97,16 +96,20 @@ class FlashCommand {
 				}
 
 				switch (info.prefixInfo.moduleFunction) {
-					case MONOLITHIC:
+					case ModuleInfo.Function.MONO_FIRMWARE:
 						// only override if modular capable
 						destSegment = specs.systemFirmwareOne ? 'systemFirmwareOne' : destSegment;
 						break;
-					case SYSTEM_MODULE:
+					case ModuleInfo.Function.SYSTEM_PART:
 						destSegment = systemModuleIndexToString[info.prefixInfo.moduleIndex];
 						destAddress = '0x0' + info.prefixInfo.moduleStartAddy;
 						break;
-					case APPLICATION_MODULE:
+					case ModuleInfo.Function.USER_PART:
 						// use existing destSegment for userFirmware/factoryReset
+						break;
+					case ModuleInfo.Function.RADIO_STACK:
+						destSegment = 'radioStack';
+						destAddress = '0x0' + info.prefixInfo.moduleStartAddy;
 						break;
 					default:
 						if (!force) {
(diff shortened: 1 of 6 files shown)


src/cmd/flash.js
```diff
@@ -97,25 +96,35 @@ class FlashCommand {
...
 				}
+
+				if (info.prefixInfo.moduleFlags & ModuleInfo.Flags.DROP_MODULE_INFO) {
+					return this._dropModuleInfo(binary);
+				}
+
+				return binary;
 			});
-		}).then(() => {
+		}).then((binary) => {
```
monkbroc: There is already a local variable `binary` so this one shadows the earlier one. Can you rename this one to `finalBinary`?

src/cmd/flash.js
```diff
@@ -135,6 +144,18 @@ class FlashCommand {
 			throw new VError(ensureError(err), 'Error writing firmware');
 		});
 	}
+
+	_dropModuleInfo(binary) {
+		// Creates a temporary binary with module info stripped out and returns the path to it
+
+		return when.promise((resolve, reject) => {
+			const rStream = fs.createReadStream(binary, { start: ModuleInfo.HEADER_SIZE });
```
monkbroc: Can you speak of the decision to stripping the module header in the CLI instead of doing it in Device OS? This is going to make it difficult to flash this file with raw dfu-util without using the CLI. I'm not opposed to the idea of doing this in the CLI, I'm just wondering what the tradeoff was.

avtolstoy: The DeviceOS handles that when the binary is flashed over serial or OTA, it's impossible to do over DFU, because all `dfu-util` does is writes a binary to a particular location and that's it and SoftDevice needs to be at a specific location without any kind of prefix (module header in our case).

avtolstoy: We could alternatively do something like putting the header at the end of the file before the footer, updating heuristics to find it (https://github.com/particle-iot/binary-version-reader/blob/master/lib/HalModuleParser.js#L202), and also updating the DeviceOS to support that, but it honestly felt like a hack.

monkbroc: Thanks for the explanation :shipit:

PR #496, opened 2019-06-28
Update ascender and Device OS for 1.1.1

technobly:
- Updates bootloader on Photon and P1 to Device OS v1.1.1's latest version (v301), even though (v201) is only required.
- Also updates all Device OS binaries to v1.1.1


technobly:
I tested this locally updating Photon/P1/Electron and A SoM.

PR #497, opened 2019-07-11
Update for Device OS for 1.2.1

technobly:
- Updated binaries for Device OS 1.2.1
- Removed Ascender mechanism for updating bootloaders and replaced with shiny new 🌟 feature implemented in [Device OS PR 1788](https://github.com/particle-iot/device-os/pull/1788) that allows us to flash the bootloader via DFU with special sequence.
- Tested on all 9 modular platforms


src/lib/deviceSpecs/specifications.js
```diff
@@ -479,6 +514,15 @@ const specs = {
 			address: '0x00030000',
 			alt: '0'
 		},
+		otaRegion: {
+			address: '0x80289000',
+			alt: '2'
+		},
+		otaFlag: {
```
m-mcgowan: This might seem like a nit-pick, but repeated data like this is possibly best factored out into an instance that is reused.

settings.js
```diff
@@ -86,38 +86,55 @@ var settings = {
...
-			systemFirmwareOne: 'argon-system-part1@1.1.1.bin'
+			systemFirmwareOne: 'argon-system-part1@1.2.1.bin',
+			otaRegion: 'argon-bootloader@1.2.1.bin',
+			otaFlag: 'ota-flag-a5.bin'
 		},
 		'2b04:d00d': {
-			systemFirmwareOne: 'boron-system-part1@1.1.1.bin'
+			systemFirmwareOne: 'boron-system-part1@1.2.1.bin',
+			otaRegion: 'boron-bootloader@1.2.1.bin',
+			otaFlag: 'ota-flag-a5.bin'
```
m-mcgowan: This was smart! And avoids lots of unnecessary file cleanup logic! 👏

PR #499, opened 2019-07-16
Fixes issue #498 where `particle update` fails on Photon/P1

### LINKED ISSUE #498: particle update (v1.43) is failing to complete dfu, new bootloader method presumed at fault
I posted an issue against device-os as it seems to be a combination of 1.2.1 new bootloader process and particle-cli 'particle update'

https://github.com/particle-iot/device-os/issues/1856

**Note the 'cannot read property 'code' of undefined.**

Marks-2017MBP:sdb-firmwareflash markterrill$ particle update

Your device is ready for a system update.
This process should take about 30 seconds. Here it goes!

! An error occurred while attempting to update the system firmware of your device:

dfu-util:
File too short for DFU suffix
dfu-util: A valid DFU suffix will be required in a future dfu-util release!!!

dfu-util:
Error during download get_status

Please visit our community forums for help with this error:
https://community.particle.io/
Cannot read property 'code' of undefined

technobly:
## Problem

When updating system firmware with `particle update` with CLI 1.43.0 on Photon/P1, the device locks up at the last step where DCT is written to.

## Solution

**Short term solution:** Need to flash Bootloader and OTA Flag before system parts for Photon/P1 because as soon as the system parts are flashed they will change the location of DFU read/write functions which live in system firmware.

**Long term solution:** Add a dirty flag in Device OS that would cause the bootloader to reload DCT functions from system parts if their flash area has been modified.  Will be added to a future Device OS release.

## Reference

Fixes #498 
https://community.particle.io/t/device-os-1-2-1-and-particle-cli-1-43-fail-in-particle-update/51140

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -86,16 +86,22 @@ var settings = {
 	},
 	updates: {
 		'2b04:d006': {
-			systemFirmwareOne: 'photon-system-part1@1.2.1.bin',
-			systemFirmwareTwo: 'photon-system-part2@1.2.1.bin',
+			// Need to flash Bootloader and OTA Flag before system parts for Photon/P1
+			// because as soon as the system parts are flashed they will change the location
+			// of DFU read/write functions which live in system firmware.
 			otaRegion: 'photon-bootloader@1.2.1.bin',
-			otaFlag: 'ota-flag-a5.bin'
+			otaFlag: 'ota-flag-a5.bin',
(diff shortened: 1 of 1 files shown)


markterrill:
Hi guys, I think there may be issues again. https://github.com/particle-iot/particle-cli/issues/498#issuecomment-558873297

PR #511, opened 2019-08-12
Also include .hpp, .hxx and .hh

nikeee:
Currently, `particle-cli` only supports `.h` files. For C++, [there is a common convention](https://stackoverflow.com/a/154445) that `.hpp`, `.hxx` or `.hh` can also be used for header files.

This PR allows `.hpp`, `.hxx` and `.hh` as  header file extensions.


ScruffR:
Would be good if we also got some means to also allow "arbitrary" files/extensions - e.g. via a `particle.include` file (much like the previously accepted `spark.ignore` file in pre `libraries 2.0` versions of CLI)

busticated:
thanks @nikeee :+1:

it'd be nice to have companion tests to verify the expected behavior. perhaps something can be easily added [here](https://github.com/particle-iot/particle-cli/blob/master/src/cmd/cloud.test.js) or as an end-to-end test over [here](https://github.com/particle-iot/particle-cli/blob/master/test/e2e/compile.e2e.js). end-to-end test docs are over [here](https://github.com/particle-iot/particle-cli/blob/master/test/README.md)

busticated:
@nikeee not sure what's up w/ CI - i tried re-running without any luck then i pulled down your branch and tests pass locally 🤔😑

nikeee:
I added an E2E test in 197f83c. However, I could not verify that it actually passes. Also, CI apperantly does not seem run the E2E tests...?

busticated:
@nikeee 

> I added an E2E test in 197f83c. 

:tada: awesome - very much appreciated :pray:

> However, I could not verify that it actually passes

are you not able to run them locally? if so, i'd be interested in learning where things went wrong so future contributors will have an easier time.

> Also, CI apperantly does not seem run the E2E tests...?

they run _after_ the unit + integration tests so if you have a failure there (which you do - for some as yet unknown reason) then the e2e tests will never run.

i would say don't worry about wrestling with it for now - what you've provided is super-helpful. i'll get it merged ASAP :+1:

busticated:
@nikeee do you mind signing our [Contributor License Agreement](https://github.com/particle-iot/particle-cli/blob/master/CONTRIBUTING.md)? for some reason i thought you had already done that 😊

nikeee:
Done.

PR #486, opened 2019-05-13
Bump serialport to 7.1.5 for Node.js 12 compability

nikeee:
Currently, I'm not able to install the current `particle-cli` due to compilation errors of `node-serialport`.

`node-serialport` fixed Node.js 12 compability in version `7.1.5`, see:
https://github.com/serialport/node-serialport/compare/serialport@7.1.4...serialport@7.1.5

PR #513, opened 2019-08-13
Add normalizeText to handle different formattings of the help text

nikeee:
As described here: https://github.com/particle-iot/particle-cli/pull/486#issuecomment-495353964
There are issues with help text. On certain CI builds, the terminal is narrower, resulting in a wrapped help text. This causes assertions regarding the help text to fail.
This PR adds a function function which strips line breaks and replaces (spaced) indentation with a single space.


nikeee:
Turns out, normalizing doesn't suffice, since the wrapped text also contains `[string]`

Issue #372, opened 2017-06-02
cannot kill the serial monitor in follow mode with CTRL-C


dmiddlecamp:
If the follow mode has disconnected / reconnected to a device after it resets, you can no longer kill the serial monitor with CTRL-C / keyboard interrupts.

technobly:
I've seen this before and when that happens CTRL+Z seemed to work.  I just tried reproducing this though and could not, perhaps this is fixed?
```
Opening serial monitor for com port: "/dev/tty.usbmodem142421"
Serial monitor opened successfully:
Serial connection closed.  Attempting to reconnect...
Serial monitor opened successfully:
// CTRL+C
^CCaught Interrupt.  Cleaning up.
Serial connection closed.
```

PR #515, opened 2019-08-21
moves hardcoded 5ms timeout to settings file

dmiddlecamp:
Background:

     We're investigating a CLI crash on a long-running `particle serial monitor --follow` invocation that runs out of memory after ~4 hours or so.

Theory:

    The follow command has a 5ms timeout before it checks for serial ports again, heavily taxing the operating system.  By making this a setting, we can more easily test a longer timeout without changing behavior for everyone.  We suspect a 50, or 100ms timeout will be easier on the host OS, and provide equivalent behavior.

Thanks!

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -47,7 +47,7 @@ var settings = {
 
 	wirelessSetupFilter: /^Photon-.*$/,
 
-	serial_follow_check_frequency: 5,
+	serial_follow_delay: 5,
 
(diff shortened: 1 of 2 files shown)


dmiddlecamp:
apologies, submitted this too early, still checking

technobly:
Closes https://github.com/particle-iot/particle-cli/issues/372 <3

PR #523, opened 2019-09-11
update for DeviceOS 1.4.0

avtolstoy:
## Description

- This PR updates DeviceOS binaries to release 1.4.0.
- Adds SoftDevice binaries
- Makes `update` command reuse `FlashCommand.flashDfu()` for non-bootloader modules. This is somewhat of a hack and ideally things need to be properly decoupled, but we are also somewhat pressed for time for 1.4.0 :disappointed: 

## How to Test

Run `particle update` for every platform.

N/A

## Completeness

- [ ] Docs have been updated
- [x] CI is passing


settings.js
```diff
@@ -91,57 +91,63 @@ var settings = {
...
+			systemFirmwareTwo: 'p1-system-part2@1.4.0.bin'
 		},
 		'2b04:d00a': {
 			// The bin files MUST be in this order to be flashed to the correct memory locations
-			systemFirmwareOne:   'electron-system-part2@1.3.1+debug.bin',
-			systemFirmwareTwo:   'electron-system-part3@1.3.1+debug.bin',
-			systemFirmwareThree: 'electron-system-part1@1.3.1+debug.bin'
+			systemFirmwareOne:   'electron-system-part2@1.4.0+debug.bin',
+			systemFirmwareTwo:   'electron-system-part3@1.4.0+debug.bin',
+			systemFirmwareThree: 'electron-system-part1@1.4.0+debug.bin'
```
busticated: do you know why we appear to be using a `debug` bin for the `electron` but nowhere else?

src/cmd/update.js
```diff
@@ -35,7 +38,25 @@ function doUpdate(id) {
...
+						// e.g. DROP_MODULE_INFO flag
+						return new FlashCommand().flashDfu({ binary: binary, requestLeave: leave })
+							.catch((err) => {
+								throw new Error(err);
+							});
+					} else {
+						// This is not a valid module and a simple binary, use DFU directly
+						return dfu.write(binary, part, leave);
+					}
+				})
```
busticated: what's different about `v1.4.0` that makes this necessary?

settings.js
avtolstoy: This is normal. By default during the release Electron binaries are built with `DEBUG_BUILD=y` to enable additional modem logs.

busticated: is it for troubleshooting? why not do it for `boron` as well?

src/cmd/update.js
avtolstoy: A mandatory SoftDevice (Nordic proprietary radio stack) upgrade.

The modules (with prefix and suffix) we generate out of the raw SoftDevice binaries provided by Nordic are somewhat special in that they contain a flag: `DROP_MODULE_INFO`. This flag indicates that before writing the module into its rightful location, the module prefix should be stripped out. It's handled by the DeviceOS when it receives a binary OTA or using YModem and it's also handled in `particle flash --usb` commands in CLI.

settings.js
avtolstoy: Yes. For troubleshooting. Boron doesn't have enough flash space currently to build with `DEBUG_BUILD=y` and all the logs.

src/cmd/update.js
busticated: got it - thanks 👍

PR #530, opened 2019-10-30
add timeout to `particle serial wifi` command to catch if device isn't listening

wraithan:
## Description

Previously when `particle serial wifi` was used and the device wasn't put in listening mode ahead of time, the command would appear to hang, forcing the user to ctrl+c or otherwise kill the process.

This patch adds the missing timeout, and adds the first tests around the `SerialCommand`'s wifi handling. 

## How to Test

`src/cmd/serial.test.js` contains automated testing for this.

### Manual testing instructions:

1. Have no devices plugged into your machine.
2. Plug in device with wifi such as a Photon and wait for it to boot to breathing cyan
3. Run `particle serial wifi` and answer no to the scan question.
4. Without patch: this would wait forever, with patch after 5s you should get a message suggesting you try `particle usb start-listening`.

This is related to, but does not solve this clubhouse ticket: https://app.clubhouse.io/particle/story/37086

## Completeness

- [ ] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/serial.js
+++ b/src/cmd/serial.js
@@ -929,7 +929,11 @@ module.exports = class SerialCommand {
 			.then(() => {
 				console.log('Done! Your device should now restart.');
 			}, (err) => {
-				log.error('Something went wrong:', err);
+				if (err && err.message) {
+					log.error('Something went wrong:', err.message);
+				} else {
+					log.error('Something went wrong:', err);
+				}
 			});
 	}
 
--- a/src/cmd/serial.js
+++ b/src/cmd/serial.js
@@ -23,13 +23,14 @@ const ensureError = require('../lib/utilities').ensureError;
 const cmd = path.basename(process.argv[1]);
 const arrow = chalk.green('>');
 const alert = chalk.yellow('!');
-const timeoutError = 'Serial timed out';
 
 function protip(){
 	const args = Array.prototype.slice.call(arguments);
 	args.unshift(chalk.cyan('!'), chalk.bold.white('PROTIP:'));
 	console.log.apply(null, args);
 }
+
+const TIMEOUT_ERROR_MESSAGE = 'Serial timed out';
 
 // An LTE device may take up to 18 seconds to power up the modem
 const MODULE_INFO_COMMAND_TIMEOUT = 20000;
@@ -616,7 +617,7 @@ module.exports = class SerialCommand {
 				return !!matches;
 			})
 			.catch((err) => {
-				if (err !== timeoutError){
+				if (err !== TIMEOUT_ERROR_MESSAGE){
 					throw err;
 				}
 				return false;
@@ -897,7 +898,7 @@ module.exports = class SerialCommand {
 			}
 
 			function startTimeout(to){
-				self._serialTimeout = setTimeout(() => reject('Serial timed out'), to);
+				self._serialTimeout = setTimeout(() => reject(TIMEOUT_ERROR_MESSAGE), to);
 			}
 
 			function resetTimeout(){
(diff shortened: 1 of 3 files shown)


src/cmd/serial.test.js
```diff
@@ -53,5 +61,21 @@ describe('Serial Command', () => {
...
+			const device = { port: 'baltimore' };
+			const mockSerial = new MockSerial();
+			serial.serialPort = mockSerial;
+			const wifiPromise = serial._serialWifiConfig(device);
+			// This allows _serialWifiConfig's internal promises to run and try to
+			// connect to the serial device before moving time forward.
+			process.nextTick(() => {
+				clock.tick(5010);
+			});
+			return expect(wifiPromise).to.be.rejected;
```
wraithan: I should change this to `rejectedWith()` and provide the error string expected. Felt bad to hardcode it in two spots though. Would love a better suggestion for that.

test/__mocks__/serial.mock.js
```diff
@@ -15,7 +15,7 @@ MockSerial.prototype.write = function write(chunk) {
 	this.data += chunk;
 };
 
-MockSerial.prototype.drain = function drain(cb) {
+MockSerial.prototype.drain = function drain(cb = () => {}) {
```
wraithan: This changes the mock interface to better match the actual interface. I considered whether that's better or if we should fix `_serialWifiConfig` to check for that error case, I did a quick check as to expected errors for `SerialPort.drain`, custom bindings and some of the low level syscalls can fail. Removing this default would mean anything tested would be forced to at least pass that callback.

src/cmd/serial.js
```diff
@@ -919,8 +920,22 @@ module.exports = class SerialCommand {
 		return promise.finally(cleanUpFn);
 	}
 
+	/**
+	 * This is a wrapper function created so _serialWifiConfig can return the
+	 * true promise state, but the wrapper can still act like it handled any
+	 * failures gracefully as it acted before testing was attempted.
+	 */
```
wraithan: This comment is probably not valuable long term, I should probably remove it and/or reword it.

src/cmd/serial.js
```diff
@@ -897,7 +898,7 @@ module.exports = class SerialCommand {
 			}
 
 			function startTimeout(to){
-				self._serialTimeout = setTimeout(() => reject('Serial timed out'), to);
+				self._serialTimeout = setTimeout(() => reject(timeoutError), to);
```
busticated: one of the many clean-up tasks we have is to start using real errors when rejecting

src/cmd/serial.js
```diff
@@ -919,8 +920,22 @@ module.exports = class SerialCommand {
 		return promise.finally(cleanUpFn);
 	}
 
+	/**
+	 * This is a wrapper function created so _serialWifiConfig can return the
+	 * true promise state, but the wrapper can still act like it handled any
+	 * failures gracefully as it acted before testing was attempted.
+	 */
+	serialWifiConfig(...args) {
+		return this._serialWifiConfig.apply(this, args)
```
busticated: prefer the simpler form 👉 `this._serialWifiConfig(...args)`

src/cmd/serial.test.js
```diff
@@ -53,5 +61,21 @@ describe('Serial Command', () => {
 			});
 		});
 	});
+
+	describe('serialWifiConfig', () => {
+		it('can reject with timeout after 5000ms if not getting any serial data', () => {
+			clock = sinon.useFakeTimers();
```
busticated: lol i always forget about `useFakeTimers()` 🤦‍♂ 

there are probably a few tests bouncing around that could benefit from that.

src/cmd/serial.test.js
busticated: yeah, errors are a bit of a mess at the moment. my aim (as noted above) is to start using _actual_ error types w/ codes and so forth. i think given the state of things what you have is fine. if anything, i'd assert against the string but w/e :+1:

not something actionable but fwiw i try to stay away from `chai` and `sinon` extensions - i've just found them more trouble than they're worth. hopefully soon (🤞😬🤞) we'll be able to use more modern js where i've found this pattern works pretty well:

```js
it('does the bar', async () => {
    let error;

    try {
        await foo.bar();
    } catch (e){
        error = e;
    }

    expect(error).to.be.an.instanceof(Error);
    expect(error).to.have.property('message', 'forgot to foo');
    // etc...
});
```

src/cmd/serial.js
wraithan: 😳 

I'm looking at this code wondering what was going through my head when I wrote it... why would you bind `this` when calling a function on `this`? Probably refactored it into that state and never simplified.

PR #542, opened 2019-12-18
bugfix/keys-quote-spaces

busticated:
## Description

Calling `particle keys doctor <device>` on a machine that uses a temporary directory path which includes spaces - e.g. `C:\path\to my temp\dir" causes error

https://app.clubhouse.io/particle/story/41811/cli-fails-key-commands-when-user-has-spaces-in-name

## How to Test

1. create a new directory with a space in its name: `mkdir /test temp dir`
2. temporarily override your temp directory env var: `TMPDIR="/test temp dir"`
3. connect your test device
4. run `particle keys doctor <device>`

the command should succeed, no errors should be reported 

## Completeness

- [x] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/src/cmd/keys.js
+++ b/src/cmd/keys.js
@@ -68,21 +68,25 @@ module.exports = class KeysCommand {
 	}
 
 	makeKeyOpenSSL(filename, alg, { protocol }) {
-		filename = utilities.filenameNoExt(filename);
+		const { filenameNoExt, deferredChildProcess } = utilities;
 
+		filename = filenameNoExt(filename);
 		alg = alg || this._getPrivateKeyAlgorithm({ protocol });
 
-		return Promise.resolve().then(() => {
-			if (alg === 'rsa') {
-				return utilities.deferredChildProcess(`openssl genrsa -out ${filename}.pem 1024`);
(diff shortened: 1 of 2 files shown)


src/cmd/keys.js
```diff
@@ -217,27 +238,26 @@ module.exports = class KeysCommand {
...
 		let pubKey = temp.path({ suffix: '.pub.pem' });
 		let inform = path.extname(filename).toLowerCase() === '.der' ? 'DER' : 'PEM';
-
 		const cleanup = () => fs.unlinkSync(pubKey);
 
 		return Promise.resolve()
 			.then(() => {
 				// try both private and public versions and both algorithms
-				return utilities.deferredChildProcess('openssl ' + algorithm + ' -inform ' + inform + ' -in ' + filename + ' -pubout -outform PEM -out ' + pubKey)
+				return deferredChildProcess(`openssl ${algorithm} -inform ${inform} -in "${filename}" -pubout -outform PEM -out "${pubKey}"`)
```
wraithan: There was a period of time where I didn't like template strings, I was very silly back then and this is way nicer to read, thanks for updating.

PR #546, opened 2020-01-31
Display info about product firmware in binary inspect command

monkbroc:
## Description

Product firmware binaries contain the product id and product firmware version baked into the binary. The `particle binary inspect` command is extended to show these values if available.

## How to Test

See automated test for example output.

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing


src/cmd/binary.js
```diff
@@ -111,6 +111,13 @@ class BinaryCommand {
 			+ ' at version '
 			+ chalk.bold(fileInfo.prefixInfo.moduleVersion.toString()));
 
+		if (fileInfo.suffixInfo.productId !== 65535 && fileInfo.suffixInfo.productVersion !== 65535) {
```
busticated: why not `fileInfo.suffixInfo.productId < 65535` ?

also couldn't hurt to make it a variable to lessen confusion - e.g. `const MAX_PRODUCT_KEY = 65535;` or some such

monkbroc: I extracted the magic strings. 65535 is a magic number specified in the linker file in case no product id and product version is set in the user firmware with `PRODUCT_ID(1234)` and `PRODUCT_VERSION(42)` so comparing with equal seems appropriate. https://github.com/particle-iot/device-os/blob/e8d4828af33387aab4d105fdfb5267a6baeb052b/build/arm/linker/module_end.ld#L5

PR #547, opened 2020-02-04
Add a command to set the setup done flag

sergeuz:
## Description

This PR implements a new command that sets/clears the setup done flag on Gen 3 devices:
```
particle usb setup-done my_device
particle usb setup-done --reset my_device
```

## How to Test

```
npm run test:e2e
```

- [ch46416]

### DIFF
--- a/test/e2e/usb.e2e.js
+++ b/test/e2e/usb.e2e.js
@@ -116,7 +116,7 @@ describe('USB Commands [@device]', () => {
 		expect(subproc.exitCode).to.equal(1);
 	});
 
-	it('Sets and clears the setup done flag', async function () {
+	it.only('Sets and clears the setup done flag', async function test() {
 		await openUsbDevice();
 		if (!usbDevice.isMeshDevice) {
 			this.skip();
--- a/test/e2e/usb.e2e.js
+++ b/test/e2e/usb.e2e.js
@@ -116,7 +116,7 @@ describe('USB Commands [@device]', () => {
 		expect(subproc.exitCode).to.equal(1);
 	});
 
-	it.only('Sets and clears the setup done flag', async function test() {
+	it('Sets and clears the setup done flag', async function test() {
 		await openUsbDevice();
 		if (!usbDevice.isMeshDevice) {
 			this.skip();
--- a/test/e2e/usb.e2e.js
+++ b/test/e2e/usb.e2e.js
@@ -1,4 +1,5 @@
 const capitalize = require('lodash/capitalize');
+const usb = require('particle-usb');
 const { expect } = require('../setup');
 const { delay } = require('../lib/mocha-utils');
 const cli = require('../lib/cli');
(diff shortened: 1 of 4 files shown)


test/e2e/usb.e2e.js
```diff
@@ -1,4 +1,5 @@
 const capitalize = require('lodash/capitalize');
+const usb = require('particle-usb');
```
busticated: i'd like to avoid using other mechanisms besides the CLI itself to manipulate the device. how about instead we go with one of the following options:

1. add a `particle usb inspect` command mirroring `particle serial inspect` but reporting device info like `id`, `type`, `platformId`, `firmwareVersion`, `vendorid`, `productid`, and status like `isOpen`, `isInDfuMode`, `.getDeviceMode()`, `isMeshDevice`, `hasModularFirmware`, etc.

2. extend `particle usb list` command output to include flags for `.getDeviceMode()` and `isMeshDevice` like it currently does for `isInDfuMode`. 

with either of those in place, we'd be able to confirm the `particle usb setup-done` state changes using the CLI itself.

what do you think?

test/e2e/usb.e2e.js
```diff
@@ -90,5 +115,34 @@ describe('USB Commands [@device]', () => {
 		expect(subproc.stderr).to.equal('');
 		expect(subproc.exitCode).to.equal(1);
 	});
-});
 
+	it.only('Sets and clears the setup done flag', async function test() {
```
busticated: need to remove the `.only` call here

test/e2e/usb.e2e.js
busticated: here's an example of option 2 👉 https://github.com/particle-iot/particle-cli/tree/setup_done/ch46416-with-list

sergeuz: The rationale was that simulating user actions on the device (resetting the device, making sure the LED is blinking blue, etc.) is a separate concern, unrelated to what we're testing here, and if there's a more direct way to do it, then why not? I also try to avoid relying on the system under test to validate its own results, if possible.

With that said, you have a lot more vision of how these tests should evolve, so if you feel strongly about this let's merge your changes in :+1:

test/e2e/usb.e2e.js
sergeuz: Fixed, thank you :+1:

test/e2e/usb.e2e.js
busticated: cool - sound good. i'll merge it in now :+1:

i was also thinking it would be nice if `particle usb setup-done` reset the device as well. as it is currently, you run that command and there's no visible change to your device until you reset it which is kind of confusing.

what do you think?

sergeuz: Resetting it is probably an overkill, but making it exit the listening mode sounds like a good idea :+1:

busticated: nice https://github.com/particle-iot/particle-cli/pull/547/commits/829c07b8c84bfee59db3ef937cc673cc245137fd - that works :+1:

busticated:
@sergeuz - i'll rebase this branch as well since there are some new changes in `master` :+1:

busticated:
thanks @sergeuz :pray::+1:

PR #549, opened 2020-02-07
Device ownership transfer is broken

monkbroc:
Story details: https://app.clubhouse.io/particle/story/46854

Steps to reproduce:
- `particle device add <deviceId>` for a device that is already claimed to another account
- Expect to be prompted with That device belongs to someone else. Would you like to request a transfer?


src/cmd/cloud.js
```diff
@@ -54,7 +54,7 @@ class CloudCommand {
 		return api.claimDevice(deviceId).then(() => {
 			console.log('Successfully claimed device ' + deviceId);
 		}, (err) => {
-			if (err && typeof err === 'string' && err.indexOf('That belongs to someone else.') >= 0) {
+			if (err && typeof err[0] === 'string' && err[0].indexOf('That belongs to someone else.') >= 0) {
```
busticated: how about adding a test so we don't hit the same issue down the road?

https://github.com/particle-iot/particle-cli/blob/master/test/e2e/cloud.e2e.js#L174-L185

monkbroc: Could you help me with that since the test needs to answer a prompt. It's not just checking stdout. It will also need a new environment variable for a device that is claimed to another account.

monkbroc:
Thanks for the example. I'll remember that.

PR #555, opened 2020-02-19
feature/ch42725/usb-dfu-confirm

busticated:
## Description

Currently, the `particle usb dfu` command is asynchronous - the command exits before the targeted device(s) are _known_ to be in DFU mode. This makes it hard to compose with other dependent commands like flashing ([example](https://github.com/particle-iot/workbench/blob/2c4fc50773c989effdb93b770e484f0a8b9fd034/packages/particle-toolchains/src/buildscripts/Makefile#L25-L28))

~~This PR adds an optional `--confirm` flag that when invoked causes the `particle usb dfu` command to wait until all targeted devices are confirmed to be in DFU mode.~~

edit: we updated the behavior in `particle-usb` to better align with `.enterListeningMode()`, etc

## How to Test

Using a user application which intentionally makes the device slow / unresponsive to usb commands:

```cpp
// ------------------------
// Simulate "busy" user app
// ------------------------

String name = "busy";
String version = "1.0.0";
unsigned long lastRun = 0;
int timeout = 1;

void setup(){
  Particle.variable("name", &name, STRING);
  Particle.variable("version", &version, STRING);
  Particle.function("timeout", setTimeout);
}

void loop(){
  while(millis() - lastRun < (timeout * 1000));
  lastRun = millis();
  Serial.printlnf("Ran at: %s [timeout: %isec]", Time.timeStr().c_str(), timeout);
}

int setTimeout(String sec){
  timeout = sec.toInt();
  Serial.printlnf("Set Timeout: %i", timeout);
  return timeout;
}
```

1. run `particle function call <device> timeout 10`
2. run `particle usb dfu <device>`
3. observe command exit timing and then reset your device

experiment with longer / shorter timeouts - e.g. `particle function call <device> timeout 30` and try targeting different collections of devices using the `--all` flag, additional named devices, and the default (`particle usb dfu`).

## Completeness

- [x] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2033,9 +2033,9 @@
       }
     },
     "@babel/runtime-corejs2": {
-      "version": "7.7.6",
-      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.7.6.tgz",
-      "integrity": "sha512-QYp/8xdH8iMin3pH5gtT/rUuttVfIcOhWBC3wh9Eh/qs4jEe39+3DpCDLgWXhMQgiCTOH8mrLSvQ0OHOCcox9g==",
+      "version": "7.8.4",
+      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.8.4.tgz",
+      "integrity": "sha512-7jU2FgNqNHX6yTuU/Dr/vH5/O8eVL9U85MG5aDw1LzGfCvvhXC1shdXfVzCQDsoY967yrAKeLujRv7l8BU+dZA==",
       "optional": true,
       "requires": {
         "core-js": "^2.6.5",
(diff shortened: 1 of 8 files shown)


busticated:
@sergeuz 

>I would prefer this to be implemented in `particle-usb` 

i agree :+1:

>Let me know if you want me to update particle-usb. It shouldn't be a big deal.

thanks. it would be most helpful if we could pair on it quickly. this way i can be better positioned to make similar tweaks in the future. toss something on my calendar :+1:

>For example, Device.enterListeningMode() does wait until the device enters the listening mode, so it would be a good opportunity to unify the library's behavior. Also, there's no need to open the device to determine if it's in the DFU mode or not, this can be determined based on the USB IDs of the attached devices.

cool! i poked around a bit over in the [repo](https://github.com/particle-iot/particle-usb) but wasn't able to get things working as i hoped. specific questions are:
* why isn't dfu considered a "mode" (e.g. `proto.DeviceMode.DFU_MODE` doesn't exist)?
* you need to open a device to put it into dfu mode, why can't we just re-query the device to determine whether or not its in dfu mode (iow, could `<device>.isInDfuMode` be made to report the current state / mode as opposed to the static value it reports today)? 
* how would you go about getting USB IDs for connected devices? is there an example within the current codebase that demonstrates that?
* how is the dfu mode / state reflected in the USB ID? or are you suggesting something like getting all devices, excluding those in dfu, and then confirming the current device is no longer in the list..?

sergeuz:
> why isn't dfu considered a "mode" (e.g. proto.DeviceMode.DFU_MODE doesn't exist)?

particle-usb's device methods are, for the most part, thin wrappers over protobuf-based control requests. Control requests are not supported in DFU mode, so there's no way a device itself can report that it's in DFU mode.

> you need to open a device to put it into dfu mode, why can't we just re-query the device to determine whether or not its in dfu mode (iow, could <device>.isInDfuMode be made to report the current state / mode as opposed to the static value it reports today)?

When a device enters DFU mode, it resets. When it resets, it detaches from the system, and the only safe operation you can do with it using the particle-usb's device instance is to close it. It cannot be reopened using the same instance as the underlying libusb handle is no longer valid.

> how would you go about getting USB IDs for connected devices? is there an example within the current codebase that demonstrates that?

There's a function that enumerates all Particle devices attached to the host: https://github.com/particle-iot/particle-usb/blob/master/src/device-base.js#L743-L751.

> are you suggesting something like getting all devices, excluding those in dfu, and then confirming the current device is no longer in the list..?

Yeah, you just need to wait until a device which successfully received a DFU request re-appears in the list in DFU mode. A better approach would be to subscribe to libusb's events which are generated when a device is attached to or detached from the system, but they're not supported in particle-usb.

busticated:
`particle-usb@0.5.4` ([release](https://github.com/particle-iot/particle-usb/releases/tag/v0.5.4)) introduces the DFU confirm behavior we need - so we'll just pull that in and skip the whole `--confirm` silliness

PR #564, opened 2020-04-08
Allow list filters to work with usb list command

JamesHagerman:
Story details: https://app.clubhouse.io/particle/story/51710

### DIFF
--- a/src/cmd/usb.js
+++ b/src/cmd/usb.js
@@ -1,5 +1,5 @@
 const { spin } = require('../app/ui');
-const { asyncMapSeries } = require('../lib/utilities');
+const { asyncMapSeries, knownPlatforms } = require('../lib/utilities');
 const { getDevice, formatDeviceInfo } = require('./device-util');
 const { getUsbDevices, openUsbDevice, openUsbDeviceById } = require('./usb-util');
 const { systemSupportsUdev, udevRulesInstalled, installUdevRules } = require('./udev');
@@ -16,6 +16,30 @@ module.exports = class UsbCommand {
 	list(args) {
 		const idsOnly = args['ids-only'];
 		const excludeDfu = args['exclude-dfu'];
+		const filter = args.params.filter;
+
+		let filterFunc = null;
+		if (filter){
+			const platforms = knownPlatforms();
+			if (filter === 'online') {
+				filterFunc = (d) => {
+					return d.connected;
+				};
+			} else if (filter === 'offline') {
+				filterFunc = (d) => {
+					return !d.connected;
+				};
+			} else if (Object.keys(platforms).indexOf(filter) >= 0) {
+				filterFunc = (d) => {
+					return d.platform_id === platforms[filter];
+				};
+			} else {
+				filterFunc = (d) => {
+					return d.id === filter || d.name === filter;
+				};
+			}
+		}
+
 		// Enumerate USB devices
 		return getUsbDevices({ dfuMode: !excludeDfu })
 			.then(usbDevices => {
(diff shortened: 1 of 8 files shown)


src/cmd/usb.js
```diff
@@ -16,6 +16,30 @@ module.exports = class UsbCommand {
...
+			} else if (Object.keys(platforms).indexOf(filter) >= 0) {
+				filterFunc = (d) => {
+					return d.platform_id === platforms[filter];
+				};
+			} else {
+				filterFunc = (d) => {
+					return d.id === filter || d.name === filter;
+				};
+			}
+		}
```
busticated, line 41: instead of copy-pasta ([source](https://github.com/particle-iot/particle-cli/blob/2c1cc519fe1b255c88bcbdb61241844249234d30/src/cmd/cloud.js#L497-L518)), let's extract a utility fn :+1:

technobly:
Would be good to add some notes in this PR for how this new feature works.  Or a link to a Docs PR.

src/lib/utilities.js
```diff
@@ -351,6 +351,39 @@ module.exports = {
...
+	 */
+	buildDeviceFilter(filter) {
+		const { knownPlatforms } = module.exports;
+		let filterFunc = null;
+		if (filter){
+			const platforms = knownPlatforms();
+			if (filter === 'online') {
+				filterFunc = (d) => {
+					return d.connected;
+				};
```
busticated, line 369: minor nit - you can collapse these down to the concise form:

```js
filterFunc = (d) => d.connected;
```

arrow fns have an implicit return ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Function_body))

test/e2e/cloud.e2e.js
```diff
@@ -74,14 +74,121 @@ describe('Cloud Commands [@device]', () => {
...
+		describe('Filter by offline', () => {
+			it('Does not show online devices', async () => {
+				args.push('offline');
+				const { stdout, stderr, exitCode } = await cli.run(args);
+
+				expect(stdout).to.not.include(`${DEVICE_NAME} [${DEVICE_ID}] (${platform})`);
+				expect(stderr).to.equal('');
+				expect(exitCode).to.equal(0);
+			});
+		});
```
busticated, line 190: nice :clap::clap::clap::+1:

i'll probably tune the naming / grouping of these a bit to better align with what i've been doing elsewhere ([example](https://github.com/particle-iot/particle-cli/blob/42091eee5bcb6779c163a30434ec9d9d83b14b0a/test/e2e/keys.e2e.js#L277-L320)) but i don't think that needs to block merging. fwiw, it matters b/c at some point i think we're going to need to be more granular in our categorization of tests (`@device` tag and related) and having them clearly grouped by command makes that more practical (setup / teardown is super-expensive atm).

src/lib/utilities.js
JamesHagerman: Indeed! Good call out. I'm far too comfortable with the function block format and overlook this refactor option _all of the time_.

test/e2e/cloud.e2e.js
JamesHagerman: Ah, yeah... Those tests are slow and the categorization is a big reason why.

Honestly, the categorization of those tests bit me here. The `Names a device` test doesn't clean itself up and expects to be ran last. Adding my `describe` block broke that assumption and required me to call `await cli.revertDeviceName();`. I'm sure there's a better way to handle that cleanup...

JamesHagerman:
@technobly For the record, usage is documented in the output of `particle usb list --help`. It's the same usage as `particle list`!

`Param filter can be: online, offline, a platform name (photon, electron, etc), a device ID or name`

PR #565, opened 2020-04-09
feature/ch47342/products-device-flash

busticated:
## Description

Adds support for optional `--product <id>` flag to the `particle cloud flash` command to facilitate flashing a product device ([ch](https://app.clubhouse.io/particle/story/47342/enable-cloud-flashing-a-product-device))

## How to Test

_Note: requires you to have a Particle product with one or more devices_

Run `particle cloud flash --help` and explore the command and options.

## Completeness

- [x] User is totes amazing for contributing!
- [x] Contributor has signed [CLA](https://docs.google.com/a/particle.io/forms/d/1_2P-vRKGUFg5bmpcKLHO_qNZWGi5HKYnfrrkd-sbZoA/viewform)
- [x] Problem and solution clearly stated
- [x] Tests have been provided
- [x] Docs have been updated
- [x] CI is passing

PR #566, opened 2020-04-10
Allow DFU flashing devices by Device ID

JamesHagerman:
Story details: https://app.clubhouse.io/particle/story/51961

Yet another CLI PR I should write tests for!

### DIFF
--- a/src/cli/keys.js
+++ b/src/cli/keys.js
@@ -31,6 +31,9 @@ module.exports = ({ commandProcessor, root }) => {
 				boolean: true,
 				default: false,
 				description: 'Force overwriting of <filename> if it exists',
+			},
+			'device': {
+				description: 'Device ID of the target device',
 			}
 		},
 		handler: (args) => {
@@ -73,6 +76,9 @@ module.exports = ({ commandProcessor, root }) => {
 				number: true,
 				description: 'Port number of the server to add to the key'
 			},
+			'device': {
+				description: 'Device ID of the target device',
+			},
 			'deviceType': {
 				description: 'Generate key file for the provided device type'
 			}
@@ -84,7 +90,11 @@ module.exports = ({ commandProcessor, root }) => {
 	});
 
 	commandProcessor.createCommand(keys, 'address', 'Read server configured in device server public key', {
-		options: protocolOption,
+		options: Object.assign({}, protocolOption, {
+			'device': {
+				description: 'Device ID of the target device',
+			}
+		}),
 		handler: (args) => {
 			const KeysCommand = require('../cmd/keys');
 			return new KeysCommand().readServerAddress(args);
--- a/src/cmd/flash.js
+++ b/src/cmd/flash.js
@@ -29,29 +29,29 @@ module.exports = class FlashCommand {
 		if (usb){
 			if (files.length > 0) {
 				// If both a device and a binary were provided, the binary will come in as the
-				// first element in the files array and device will be accurate
+				// first element in the files array
 				binary = files[0];
-
-				// Lookup the Device ID based on the provided Device Name
-				if (!isDeviceId(device)) {
-					const api = new ParticleApi(settings.apiUrl, { accessToken: settings.access_token }).api;
-
-					let deviceInfo
-					try {
-						deviceInfo = await getDevice({ id: device, api, auth: settings.access_token});
-					} catch (err) {
-						throw VError(ensureError(err), 'Device ID lookup failed');
-					}
-
-					if (deviceInfo && deviceInfo.id) {
-						device = deviceInfo.id;
-					} else {
-						throw VError('Device ID lookup failed');
-					}
-				}
 			} else {
 				// Otherwise, no device has been provided so unset the inaccurate value
 				device = undefined;
+			}
+
+			// Lookup the Device ID based on the provided Device Name
+			if (device && !isDeviceId(device)) {
+				const api = new ParticleApi(settings.apiUrl, { accessToken: settings.access_token }).api;
+
+				let deviceInfo;
+				try {
+					deviceInfo = await getDevice({ id: device, api, auth: settings.access_token });
+				} catch (err) {
+					throw new VError(ensureError(err), 'Device ID lookup failed');
+				}
+
+				if (deviceInfo && deviceInfo.id) {
+					device = deviceInfo.id;
+				} else {
+					throw new VError('Device ID lookup failed');
+				}
 			}
 
 			result = this.flashDfu({ device, binary, factory, force });
@@ -83,7 +83,7 @@ module.exports = class FlashCommand {
 		return Promise.resolve()
 			.then(() => dfu.isDfuUtilInstalled())
 			.then(async () => {
-				targetDevice = await dfu.findCompatibleDFU({ deviceId: device })
+				targetDevice = await dfu.findCompatibleDFU({ deviceId: device });
 			})
 			.then(() => {
 				//only match against knownApp if file is not found
--- a/src/cmd/flash.js
+++ b/src/cmd/flash.js
@@ -1,6 +1,9 @@
 const fs = require('fs');
 const VError = require('verror');
 const dfu = require('../lib/dfu');
+const settings = require('../../settings');
+const ParticleApi = require('./api');
+const { getDevice, isDeviceId } = require('./device-util');
 const ModuleParser = require('binary-version-reader').HalModuleParser;
 const ModuleInfo = require('binary-version-reader').ModuleInfo;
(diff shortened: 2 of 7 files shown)


src/cmd/flash.js
```diff
@@ -24,16 +27,42 @@ module.exports = class FlashCommand {
 
 		let result;
 		if (usb){
-			result = this.flashDfu({ binary, factory, force });
+			if (files.length > 0) {
+				// If both a device and a binary were provided, the binary will come in as the
+				// first element in the files array
+				binary = files[0];
```
busticated, line 33: this is ambiguous at best and likely inaccurate. i've been adding tests like [this](https://github.com/particle-iot/particle-cli/pull/565/commits/33595ec6fc136038318562fa7d6e805dab817e47) to help clarify similar behavior.

src/cmd/flash.js
```diff
@@ -24,16 +27,42 @@ module.exports = class FlashCommand {
...
+			} else {
+				// Otherwise, no device has been provided so unset the inaccurate value
+				device = undefined;
+			}
+
+			// Lookup the Device ID based on the provided Device Name
+			if (device && !isDeviceId(device)) {
+				const api = new ParticleApi(settings.apiUrl, { accessToken: settings.access_token }).api;
```
busticated, line 41: at some point i think we're going to have to insist of device IDs across the board as this type of thing leads to confusion - e.g. "why am i getting 4xx errors when flashing locally?!" etc

technobly:
Would be good to add some notes in this PR for how this new feature works.  Or a link to a Docs PR.

monkbroc:
The 2 relevant CLI docs pages are:
https://docs.particle.io/reference/developer-tools/cli/
https://docs.particle.io/tutorials/developer-tools/cli/

JamesHagerman:
I've got a lot more work to do on this before it's ready to merge. I'll make this PR a draft until I'm closer to the mark.

src/cli/keys.js
```diff
@@ -73,6 +76,9 @@ module.exports = ({ commandProcessor, root }) => {
 				number: true,
 				description: 'Port number of the server to add to the key'
 			},
+			'device': {
+				description: 'Device ID of the target device',
+			},
```
busticated, line 81: we have a number of commands in this same namespace that accept `deviceID` as a _parameter_ - e.g. https://github.com/particle-iot/particle-cli/blob/5659edf2001163fa4f88a3db6057c91276e74e9a/src/cli/keys.js#L56-L63 

i think we should either support that call signature (total pain, requires manual arg checking / juggling, etc) - OR - we rename the option (flag) to `--deviceID`. in both cases we want to be more explicit about the expected input (since `device` means `id` or `name` across the rest of the CLI commands)

src/cmd/flash.js
```diff
@@ -24,16 +27,42 @@ module.exports = class FlashCommand {
...
+
+				let deviceInfo;
+				try {
+					deviceInfo = await getDevice({ id: device, api, auth: settings.access_token });
```
busticated, line 45: `getDevice()` from `device-util` doesn't really do anything special - i'd love it if we could just use the `ParticleApi` bits directly so eventually i can just remove `device-util` entirely.

src/lib/dfu.js
```diff
@@ -97,42 +118,74 @@ module.exports = {
...
+									dfuId: ans.device.dfuId,
+									deviceId: ans.device.deviceId
 								};
 							});
-						}
-					};
-					return prompt([question])
-						.then((ans) => {
-							const dfuId = ans.device;
-							module.exports.dfuId = dfuId;
```
busticated, line 122: hm. ok. yeah.

wheee!!

busticated: (for clarity, i'm referring to our `exports` usage here not that this line changed 🤗)

JamesHagerman: Ha! Yeah. Not sure I've ever seen `module.exports` used quite like that. I guess that's one way to force private members?

In any case, it makes refactoring DFU related code tricky.

JamesHagerman:
Sorry, this is a bit of a rant, but I've been working this over in my head for a bit and need to put my thoughts somewhere...

> to that end, i'm thinking we should take a step back and think about how this all could be simplified - e.g. _always_ target by id so we don't need to worry about non `-S` usage? use `particle-usb` and related to grab device id? add some special / isolated command namespace to host this functionality? other?

Maintaining backwards compatibility with DFU is difficult. The DFU serial numbers used by older bootloaders don't match the Device ID nor are they even unique between devices. :disappointed: User intervention is required in cases where more than one older device is in DFU mode at a time.

That said, we do need a better way to host this functionality. If nothing else, we need better separation of concerns between the commands themselves and the DFU implementation. Currently, both the commands AND the command-specific implementations in the DFU module have a good portion of code duplication.

Because of that duplication, I don't think an isolated command namespace would help.

> re: testing - extending the `e2e` suite to expect _two_ devices connected via USB is probably going to be a major pain (mainly b/c other CLI commands don't handle multiple devices well) so as a first step what i'd want to see is solid testing around the happy path (one device) to ensure we aren't breaking existing functionality.

Indeed! While manually testing this, I was just about to commit when I happened to plug in three devices and found new bugs. :face_with_head_bandage:

Honestly, DFU isn't going away. While `particle-usb` might help in some areas there will always be limits on how far we can (or should) re-implement the stuff we've gotta accomplish. So, it's worth treating the DFU layer as a first class citizen at some point.

I can think

In my mind, that means it needs to be restructured. It needs to not lean so heavily on the `module.exports` hackery for data/state storage. It needs structure to reduce duplication between it's methods. It needs structure for separation of concerns between the calling code and the internal implementation. It needs structure to support testing. And we need tests before we can gracefully restructure it. Chicken, egg, etc.

I've been sitting on this because we have no immediate need for this functionality... but I think this is a really good example of how we might want to tackle other areas of the CLI. My intention is to draw a line in the sand, isolate all refactoring to the lower implementation details while also leaving as much of the existing code in place as possible since that starts quite a yak shave through a bunch of other code. My planned approach has been loosely this:

1. Attempt to convert `dfu.js` so it's in the following format:

```
let something...
module.exports = { something };
```

If nothing else, just for readability.

But more importantly, that adventure will explicitly call out the `module.exports` pattern (anti-pattern?) that we're using in each of those methods. Specifically, I'm speaking about code like this, `const { listDFUDevices, showDfuModeHelp } = module.exports;` which expects `module.exports` to work in ways I'm not sure is guaranteed in Node.

It would be a good thing to move towards using an Object (or Class) which has a scope we can explicitly manipulate for things like... I don't know, storing data? Writing software? Fixing the world?... as well as moving us towards a way to build test fixtures that don't depend on the internal workings of Node's export implementation for our tests to function.

:huff:

2. THEN, Export a new object that actually has a manageable scope (a Class, a real Object, whatever) which can sit next to the existing named exports. 
3. Create tests for that new scoped object export
4. Start moving the USB methods into that object one by one as the tests come up to speed.

That approach should (in theory) let us keep the majority of the calling code as is while we iterate on a refactored implementation/interface. As we go, it will be more apparent where the calling code has to move and whether we need to back track on our implementation to handle unexpected issues.

Obviously, this is all just noise until I sit down and open the editor, but I see a path through the trees!

busticated:
>Maintaining backwards compatibility with DFU is difficult

in what sense? can it be rectified by shipping different toolchains? versioning an interface?

> it's worth treating the DFU layer as a first class citizen at some point

agree. we simply must be able to write to our devices 100% reliably. 

>...we need better separation of concerns between the commands themselves and the DFU implementation

absolutely agree. we have this problem throughout this codebase. lots of intermingled `stdout.write()` calls. etc.

>I don't think an isolated command namespace would help

👌 

>Attempt to convert dfu.js

thanks for describing the refactors - i agree it needs some tending to but at this point you've probably stared at the code longer than i have. my big concern is simply not breaking what we have today... or if we do, that the thing we replace it with is known-good (solid tests). i'll schedule some time for us to chat.

>Specifically, I'm speaking about code like this, `const { listDFUDevices, showDfuModeHelp } = module.exports;` which expects module.exports to work in ways I'm not sure is guaranteed in Node.

with you up until this minor bit. can you elaborate? this is actually i pattern i use a lot and i'm not aware of any unmet expectations..? i was referring to setting the exports object asynchronously - that code for sure breaks in our slowly approaching esm future.

everything else, 👍 

>Obviously, this is all just noise until I sit down and open the editor, but I see a path through the trees!

nice! that counts 👍

JamesHagerman:
Closing this story as this work is stale and is not an often enough requested feature.

PR #588, opened 2020-07-09
feature/ch55094/usb-list-device-timeout-fallback

busticated:
## Description

Running `particle usb list` when one of your connected devices is "busy" or otherwise unresponsive causes the entire list operation to fail with `Request timeout`. This change handles device communication timeouts such that your device list is still available regardless.

## How to Test

Using a device running some **very bad** user firmware like:

```cpp
#include "Particle.h"

void setup() {
}

void loop() {
    while(1);
}
```

...run `particle usb list` with multiple devices connected and verify you still get your device list back

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/usb.js
+++ b/src/cmd/usb.js
@@ -1,7 +1,7 @@
 const { spin } = require('../app/ui');
 const { asyncMapSeries, buildDeviceFilter } = require('../lib/utilities');
 const { getDevice, formatDeviceInfo } = require('./device-util');
-const { getUsbDevices, openUsbDevice, openUsbDeviceById } = require('./usb-util');
+const { getUsbDevices, openUsbDevice, openUsbDeviceById, TimeoutError } = require('./usb-util');
 const { systemSupportsUdev, udevRulesInstalled, installUdevRules } = require('./udev');
 const { platformsById } = require('./constants');
 const ParticleApi = require('./api');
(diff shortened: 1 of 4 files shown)


src/cmd/usb.js
```diff
@@ -43,7 +43,15 @@ module.exports = class UsbCommand {
 							let info = [device, usbDevice.isInDfuMode];
 
 							if (!usbDevice.isInDfuMode){
-								info.push(usbDevice.getDeviceMode());
+								info.push(
+									usbDevice.getDeviceMode()
+										.catch(error => {
+											if (error.message.includes('timeout')){
```
sergeuz, line 49: While testing [particle-usb/#29](https://github.com/particle-iot/particle-usb/pull/29), I noticed that sometimes timeout exceptions thrown by the library don't have a message :disappointed: It's better to check if `error` is an instance of particle-usb's `TimeoutError`. I also added a timeout option to `getDeviceMode()` – setting it to something like 2-3 seconds will prevent the list command from taking too long if one of the devices is blocked.

busticated: > ...sometimes timeout exceptions thrown by the library don't have a message...

:joy::sob: whaa?.. but... how? isn't the timeout error coming from [here](https://github.com/particle-iot/particle-usb/blob/04837184cc80734f240eadb5fb2ef3d2c4f08e8c/src/device-base.js#L232)?

...oh maybe it's coming from [here](https://github.com/particle-iot/particle-usb/blob/master/src/device.js#L68-L88) sometimes..?

why not assign a default `message` [here](https://github.com/particle-iot/particle-usb/blob/04837184cc80734f240eadb5fb2ef3d2c4f08e8c/src/error.js#L49) - e.g. 

```js
this.message = this.message || 'Request timeout';
```

sergeuz: Yeah, there will be a message in this specific case, but methods that rely on `RequestSender` internally throw `TimeoutError` without any message set :shrug: 

> why not assign a default message

:+1:

busticated: ✅ in https://github.com/particle-iot/particle-cli/pull/588/commits/368c0584c382862c2f25a88d1407124b7bfaa274

PR #592, opened 2020-08-15
bugfix/ch61378/flash-tinker-product-device

busticated:
## Description

running the `particle flash <my-device> tinker` when `<my-device>` is in a product fails with: `Failed to flash <device-name>: Cannot read property 'map' of undefined` ([ch](https://app.clubhouse.io/particle/story/61378/particle-flash-my-device-tinker-fails-when-device-is-in-a-product))

## How to Test

using a product device, run `particle flash <my-device> tinker` then repeat that command using a non-product device and verify that both devices are flashed appropriately

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/cloud.js
+++ b/src/cmd/cloud.js
@@ -247,7 +247,22 @@ module.exports = class CloudCommand extends CLICommandBase {
 
 		return createAPI().getDeviceAttributes(deviceId)
 			.then((attrs) => {
-				const productId = product ? attrs.platform_id :attrs.product_id;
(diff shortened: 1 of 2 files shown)


busticated:
>Ouch, the conditions hurt my brain.

haha! yeah, same. 

> `platform_id` is always defined for any device so it should be preferred over `product_id`

ah, nice. that helps - i can drop the `else` in that case at least - thanks :pray::+1:

PR #607, opened 2021-03-20
bug/ch75832/particle-usb-list-platform-filter

busticated:
## Description

running the `particle usb list <platform - e.g. tracker>` command fails to return matches when you have a device of that platform connected but not associated with your account ([ch](https://app.clubhouse.io/particle/story/75832/particle-usb-list-platform-does-not-work-when-device-belong-to-an-account-other-than-the-one-you-are-signed-in-with))

## How to Test

with an `argon` device connected via USB but which is not associated with the signed-in user, run `particle usb list argon` and verify that your device is listed in the result set

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/usb.js
+++ b/src/cmd/usb.js
@@ -73,7 +73,7 @@ module.exports = class UsbCommand {
 								id: usbDevice.id,
 								name: name || '',
 								type: `${type.join(', ')}`,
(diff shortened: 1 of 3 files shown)


monkbroc (approved):
👍

PR #608, opened 2021-04-23
Add support for XDG_DATA_HOME

nrobinson2000:
## Description

In particle-cli-wrapper, the default location for particle-cli dependencies can be overridden by setting the `XDG_DATA_HOME` environment variable.

https://github.com/particle-iot/particle-cli-wrapper/blob/master/filesystem.go#L21

However, in particle-cli this override is not honored and the default path of `$HOME/.particle` is used.

In particle-cli-wrapper, when `XDG_DATA_HOME` is set, the path becomes `$XDG_DATA_HOME/particle`. With the changes I've proposed so far, the path for particle-cli would also become `$XDG_DATA_HOME/particle`.

## How to Test

- Copy the `particle` executable from particle-cli-wrapper or Workbench to the current working directory.

- Run the following commands:

```bash
$ rm -rf ~/.particle
$ mkdir build
$ export XDG_DATA_HOME="$PWD/build"
$ ./particle
```

- Observe the contents of `build`:

```text
$ tree -L 2 build
build
`-- particle
    |-- autoupdate
    |-- error.log
    |-- node-v12.16.1-linux-x64
    |-- node_modules
    |-- package-lock.json
    |-- plugin-cache.json
    `-- tmp

4 directories, 4 files
```

- Observe that `~/.particle` was created:

```bash
$ ls ~/.particle
profile.json
```

With the changes I'm proposing, `profile.json` would have been created at `build/particle/profile.json` and not `~/.particle/profile.json`.

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -201,7 +201,12 @@ settings.findHomePath = () => {
 };
 
 settings.ensureFolder = () => {
-	let particleDir = path.join(settings.findHomePath(), '.particle');
+	let base = '.particle';
+	let dir = process.env['XDG_DATA_HOME'];
+	if (dir && fs.existsSync(dir)) {
+		base = 'particle';
+	}
+	let particleDir = path.join(settings.findHomePath(), base);
 	if (!fs.existsSync(particleDir)) {
 		fs.mkdirSync(particleDir);
 	}
--- a/settings.js
+++ b/settings.js
@@ -184,6 +184,7 @@ function envValueBoolean(varName, defaultValue) {
 
 settings.findHomePath = () => {
 	let envVars = [
+		'XDG_DATA_HOME',
 		'home',
 		'HOME',
 		'HOMEPATH',


nrobinson2000:
@busticated, does this look good to you?

settings.js
```diff
@@ -184,6 +184,7 @@ function envValueBoolean(varName, defaultValue) {
 
 settings.findHomePath = () => {
 	let envVars = [
+		'XDG_DATA_HOME',
```
busticated, line 187: `$HOME` and `$XDG_DATA_HOME` are not equivalent.

from the [spec](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html):

> $XDG_DATA_HOME defines the base directory relative to which user-specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

busticated (commented):
thanks for the contribution 🙏 

unfortunately, we have a number of tools and processes (most internal to Particle) that depend on the current behavior / file locations and which cannot easily be updated to accommodate this new logic. 

i'm going to close this out but will keep it in mind for future major updates here and beyond.

nrobinson2000:
Hey, @busticated 

When using particle-cli/particle-cli-wrapper I think it's useful to have the ability to specify an alternate location for dependencies. All this pull request does is make `particle-cli` match the behavior of `particle-cli-wrapper`, which is to use `$XDG_DATA_HOME/particle` only if `XDG_DATA_HOME` has explicitly been set. Otherwise, it defaults to `$HOME/.particle`. **If `XDG_DATA_HOME` is not set, the behavior is unchanged.**

In particle-cli there is a list of [environment variables to consider](https://github.com/nrobinson2000/particle-cli/blob/patch-2/settings.js#L186) and `HOME` takes medium priority. As shown in the list, `home` has greater priority than `HOME`.

I can export it and run `particle`:

```shell
$ mkdir build
$ export home="$PWD/build"
$ particle
```

Afterwards, `.particle/profile.json` is created:

```shell
$ tree -a build
build
`-- .particle
    `-- profile.json

1 directory, 1 file
```

Are there any situations where an (internal) tool or process dependent on `particle-cli` utilizes `XDG_DATA_HOME` but still expects `particle-cli` to use `$HOME/.particle`?

Is there an alternative environment variable you'd like to use? Or is the use of `particle` rather than `.particle` for the basename of the directory to match the behavior of `particle-cli-wrapper` the problem?

busticated:
sorry @nrobinson2000 it's just not a change we're going to make over here for now. thanks for understanding 🙏

PR #617, opened 2021-12-01
Device specs from Device Constants

monkbroc:
Story details: https://app.shortcut.com/particle/story/88683

Pull information from `@particle/device-constants` to generate the device specs used in the CLI

Changes:
- Move the known apps into a directory structure so they can be auto-populated (also remove ancient apps for Spark Core)
- Remove support for Duo
- Removes Coveralls since we don't use this service anymore

Test plan:
- [X] `particle serial inspect` shows the platform display name
- [x] `particle keys server` and `particle keys save` work
- [x] `particle flash --usb` works for user app and system part
- [x] `particle usb list` and `particle usb dfu` works
- [x] E2E tests run (I ran `npm run test:e2e -- --grep "product device" --invert` because I didn't have a product device set up)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4782,19 +4782,19 @@
       "integrity": "sha1-+8cfDEGt6zf5bFd60e1C2P2sypE="
     },
     "form-data": {
-      "version": "2.3.3",
-      "resolved": "https://registry.npmjs.org/form-data/-/form-data-2.3.3.tgz",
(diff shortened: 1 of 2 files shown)


monkbroc:
Most E2E tests passed. The only failure that needs investigation is
```
    USB Setup-Done Subcommand
      39) Sets and clears the setup done flag
```

I fixed it in the latest commit.

busticated (approved):
🙌 💪 🙏 👍

PR #641, opened 2023-05-18
Feature/sc 118328/compile source before bundle

keeramis:
## Description

Adds support to bundle assets and application binary from the project folder

### Usage

```
npm start -- compile argon /path/to/project/folder/with/nested/assets/folder --saveTo appBundle.zip

```

### Example
```
particle-cli (feature/sc-118328/compile-source-before-bundle) $ npm start -- compile argon /Users/ka/code/workbench/argon-tests --saveTo appBundle.zip

> particle-cli@3.10.1 start
> node ./src/index.js compile argon /Users/ka/code/workbench/argon-tests --saveTo appBundle.zip


Compiling code for argon

Including:
    ../workbench/uhura-tests/src/uhura-tests.cpp
    ../workbench/uhura-tests/project.properties
    /Users/keerthyamisagadda/code/workbench/uhura-tests/assets/IMG-20230507-WA0004.jpg
    /Users/keerthyamisagadda/code/workbench/uhura-tests/assets/PXL_20230423_091313262.jpg

attempting to compile firmware
downloading binary from: /v1/binaries/6467e2ff0b281823c47f28c3
Memory use:
   text	   data	    bss	    dec	    hex	filename
   3744	    112	    276	   4132	   1024	/workspace/target/workspace.elf

Compile succeeded and bundle created.
Saved bundle to: /Users/keerthyamisagadda/code/particle-cli/appBundle.zip
```
All the other legacy use cases should be supported as-is.

## How to Test
1. Pull down the branch: `git pull && git checkout feature/asset-bundle`
2. Run `nvm use`
3. Reinstall dependencies: `npm run reinstall`
4. Run tests: `npm test`

https://app.shortcut.com/particle/story/118328/create-zipped-app-bundle-from-source-file

## Completeness

- [x] Docs have been updated
- [x] CI is passing


src/cmd/cloud.js
```diff
@@ -260,17 +261,32 @@ module.exports = class CloudCommand extends CLICommandBase {
...
 	}
 
-	_getDownloadPath(deviceType, saveTo){
-		if (saveTo){
-			return saveTo;
+	_getDownloadPathForBin(deviceType, saveTo){
+		if (saveTo) {
+			// If saveTo has a .bin extension, use it. Otherwise,
+			// remove any existing extension and append .bin to the filename.
+			return (utilities.getFilenameExt(saveTo) === '.bin') ? saveTo : (utilities.filenameNoExt(saveTo) + '.bin');
```
keeramis, line 268: Made this modification because in the case of the following,
```
npm start -- compile /path/to/proj/folder/with/assets --saveTo myBundle.zip
```
the compiled binary is also ended up being called `myBundle.zip` so changing the logic to call app binary with .bin extension. I wonder this is backwards compatible.

src/cmd/cloud.js
```diff
@@ -322,6 +338,14 @@ module.exports = class CloudCommand extends CLICommandBase {
...
 				}
 
+				const assetsPath = this._checkForAssets(files);
+				if (assetsPath) {
+					return new BundleCommands()._getAssets({ assetsPath });
+				}
+			})
+			.then((result) => {
+				// If assets are not present, assets will be undefined and any future 'if' checks will fail
+				assets = result;
```
keeramis, line 348: the `assets` variable set here will suggest if the asset mode is on or not.

src/cmd/cloud.js
monkbroc: If you flip the logic around to 
```suggestion
			return (utilities.getFilenameExt(saveTo) === '.zip') ? (utilities.filenameNoExt(saveTo) + '.bin') : saveTo;
```
then it won't unintentionally change the behavior in case someone was doing `particle compile --saveTo filename_no_extension`

src/cmd/cloud.js
```diff
@@ -260,17 +261,32 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+		if (saveTo) {
+			// If saveTo has a .bin extension, use it. Otherwise,
+			// remove any existing extension and append .bin to the filename.
+			return (utilities.getFilenameExt(saveTo) === '.bin') ? saveTo : (utilities.filenameNoExt(saveTo) + '.bin');
 		}
-
 		return deviceType + '_firmware_' + Date.now() + '.bin';
 	}
 
+	_getDownloadPathForZip(deviceType, saveTo, assets){
```
monkbroc, line 273: Call this _getBundleSavePath to match the equivalent function in bundle.js

src/cmd/cloud.js
```diff
@@ -260,17 +261,32 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+		if (!assets) {
+			return;
+		}
+		if (!saveTo){
+			return deviceType + '_firmware_' + Date.now() + '.zip';
+		} else if (path.extname(saveTo) !== '.zip'){
+			throw new Error('saveTo must have a .zip extension');
```
monkbroc, line 280:
```suggestion
			throw new Error('saveTo must have a .zip extension when project includes assets');
```

src/cmd/cloud.js
```diff
@@ -645,6 +694,24 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+	/**
+	 * Checks if the given filepath contains an assets directory
+	 * @param filePath
+	 * @returns {string} path to assets directory if it exists, otherwise undefined
+	 * @private
+	 */
+	_checkForAssets(files) {
+		for (let i = 0; i < files.length; i++) {
+			const file = files[i];
+			if (fs.statSync(file).isDirectory()) {
```
monkbroc, line 706: Add a try catch around this in case the file doesn't exist.

src/cmd/cloud.js
```diff
@@ -378,16 +412,31 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+				if (respSizeInfo){
 					this.ui.stdout.write(`Memory use:${os.EOL}`);
-					this.ui.stdout.write(`${sizeInfo}${os.EOL}`);
+					this.ui.stdout.write(`${respSizeInfo}${os.EOL}`);
+				}
+				if (bundle) {
+					if (fs.existsSync(filename)){
+						fs.unlinkSync(filename);
+					}
+					// TODO: Verify that the bundle is created with the correct files
```
monkbroc, line 430: Remember to remove this comment before merging the PR

src/cmd/cloud.js
```diff
@@ -378,16 +412,31 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+					this.ui.stdout.write(`Saved bundle to: ${path.resolve(bundleFilename)}${os.EOL}`);
+				} else {
+					this.ui.stdout.write(`Compile succeeded.${os.EOL}`);
+					this.ui.stdout.write(`Saved firmware to: ${path.resolve(filename)}${os.EOL}`);
 				}
-				this.ui.stdout.write(`Compile succeeded.${os.EOL}`);
-				this.ui.stdout.write(`Saved firmware to: ${path.resolve(filename)}${os.EOL}`);
+			})
+			.catch((err) => {
+				throw normalizedApiError(err);
```
monkbroc, line 439: normalizedApiError should only be called after a Particle API call. It's not suitable for a general catch clause after a long function. What issue did you encounter that you are trying to fix by adding this?

src/cmd/cloud.js
```diff
@@ -341,20 +365,28 @@ module.exports = class CloudCommand extends CLICommandBase {
 					for (let i = 0, n = list.length; i < n; i++){
 						this.ui.stdout.write(`    ${list[i]}${os.EOL}`);
 					}
+					if (assets) {
+						this.ui.stdout.write(`Including assets:${os.EOL}`);
+						_.values(assets).forEach((asset) => {
+							this.ui.stdout.write(`    ${asset.name}${os.EOL}`);
```
monkbroc, line 371: Let's print the path to assets as well
```suggestion
							this.ui.stdout.write(`    ${asset.path}${os.EOL}`);
```

Then you can remove the `Including assets:` line above and it will print a combined list of source code and assets. It should be clear which is which because the assets are last and start with `assets/`

Note that you'll need to update `getAssets` to set the `path` property to `path.join(assetsPath, name)`

test/e2e/compile.e2e.js
```diff
@@ -550,6 +551,120 @@ describe('Compile Commands', () => {
...
+		const { stdout, stderr, exitCode, start, end } = await cliRunWithTimer(args, { cwd });
+		const file = await fs.stat(destinationZip);
+		const log = [
+			`Compiling code for ${platform}`,
+			'',
+			'Including:',
+			'    src/stroby.ino',
+			'    project.properties',
+			'',
+			'Including:',
```
monkbroc, line 568: This will need to be updated based on my comment combining the source and assets list.

monkbroc (approved):
Nice work! I left a few comments on things to improve but it already works great!

src/cmd/cloud.js
keeramis: I was thinking along of lines of using BVR's `unpackApplicationAndAssetBundle` over here but wanted to ask for your suggestion before doing that. If it's not needed to be too specific, I will remove it.

src/cmd/cloud.js
keeramis: Thank you for highlighting. The exact scenario is this e2e test which will not return the error https://github.com/particle-iot/particle-cli/blob/master/test/e2e/compile.e2e.js#L502C29-L525

so I pulled the catch block down to the very ending. I am guessing I will have to refactor this a bit 🤔

keeramis: Refactored to use async/await so I believe my concern here doesn't hold true anymore :)

PR #646, opened 2023-06-08
Rework particle.include and particle.ignore for CLI compile

keeramis:
## Description

The CLI cloud compile command has an undocumented pair of files `particle.include` and `particle.ignore` that modify which files are sent to the cloud compiler. The implementation is broken because adding files to particle.include strips the relative path. 

This PR reworks the usage of `particle.include` and `particle.include` files such that the file pattern in those files are used to either include/remove those files from the list of files that is sent to the cloud compiler.

## How to Test

1. Setup `particle-cli` locally as per the instructions [here](https://github.com/particle-iot/particle-cli#development)
2. Checkout this branch
3. Run `npm start -- compile <platform> </path/to/project>`

### Adding `particle.include` or `particle.ignore` files

The `particle.include` and `particle.ignore` files provide a way to include or exclude _additional_ files from the cloud compiler. Please note that these files are only for special extras and the default files like the source and header files are handled separately and are not affected by the patterns specified in these two files. These files can be located either in the project's root directory or in any of its subdirectories.

When the `particle.include` file is placed in the root directory, files matching the specified patterns are searched recursively from the root directory. On the other hand, if the `particle.include` file is placed in a subdirectory, the search for matching files is performed recursively starting from that subdirectory as the base. The system automatically handles duplicates, so there is no need to worry about them.

Similarly, the `particle.ignore` file follows the same behavior, but instead of including files, it excludes files that match the specified patterns from being sent to the cloud compiler.

- Refer to the examples here for particle.include -  [Example project structure to include files](https://github.com/particle-iot/particle-cli/tree/bug/sc-119000/project-incl-files/test/__fixtures__/projects/proj-include)
- Refer to the examples here for particle.ignore -  [Example project structure to ignore files](https://github.com/particle-iot/particle-cli/tree/bug/sc-119000/project-incl-files/test/__fixtures__/projects/proj-ignore)

https://app.shortcut.com/particle/story/119000/rework-particle-include-and-particle-ignore-functionality-in-cli-compile

## Completeness

- [x] Docs have been updated
- [x] CI is passing


src/lib/utilities.js
```diff
@@ -205,6 +206,10 @@ module.exports = {
 				line = path.join(basepath, line);
 			}
 			found = glob.sync(line, { nodir: true, follow: !!followSymlinks });
+			if (os.platform() === 'win32') {
+				// because glob pattern is always in posix format
+				found = found.map(file => path.win32.normalize(file));
+			}
```
keeramis, line 212: Apparently, `glob.sync` only works with POSIX formats, which caused it to return POSIX-formatted file paths even on Windows systems. There is a `--posix: false` that theoritically should work around this problem but that option is not working as expected. This has not been a problem in our codebase because the file paths were later processed by `path.relative` [in cloud.js](https://github.com/particle-iot/particle-cli/blob/e8354b92ad54a98637ac2a54caf7d5b9c43ec833/src/cmd/cloud.js#LL823C4-L823C4), which handled the necessary adjustments for the specific operating system but when I tried to write unit tests for that specific code, it started to make the tests fail.

I believe it's best to modify the behavior of glob.sync to generate Windows-formatted paths directly at the source. This way, we can eliminate the need for the caller to handle this issue.

keeramis:
I verified Windows is OK as well! @monkbroc , and it adds the .def files even when the content is in unix format.

This was the `particle.include` file:
```
**/*.def
```
Here's the text from console.

[windows-tracker-edge-and-include.txt](https://github.com/particle-iot/particle-cli/files/11707995/windows-tracker-edge-and-include.txt)

src/cmd/cloud.js
```diff
@@ -779,11 +779,24 @@ module.exports = class CloudCommand extends CLICommandBase {
...
 	_processDirIncludes(fileMapping, dirname, { followSymlinks } = {}){
 		dirname = path.resolve(dirname);
+		let files = new Set();
 
-		const includesFile = path.join(dirname, settings.dirIncludeFilename);
-		const ignoreFile = path.join(dirname, settings.dirExcludeFilename);
-		let hasIncludeFile = false;
+		this._getDefaultIncludes(files, dirname, { followSymlinks });
+		this._getCustomIncludes(files, dirname, { followSymlinks });
+		this._getDefaultIgnores(files, dirname, { followSymlinks });
```
monkbroc, line 786: Put default ignores before custom includes so it's possible to re-include something that got ignored.

src/cmd/cloud.js
```diff
@@ -793,45 +806,57 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+		const includeFiles = utilities.globList(dirname, ['**/particle.include'], { followSymlinks });
+		const result = [];
 
+		for (const includeFile of includeFiles) {
+			const includeDir = path.dirname(includeFile);
+			const globsToInclude = utilities.trimBlankLinesAndComments(utilities.readAndTrimLines(includeFile));
+			if (!globsToInclude || !globsToInclude.length) {
+				continue;
+			}
+			const globList = globsToInclude.map(g => g);
```
monkbroc, line 827: What is thing line for?

src/cmd/cloud.js
```diff
@@ -793,45 +806,57 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+			const includePaths = utilities.globList(includeDir, globList, { followSymlinks });
+			result.push(...includePaths);
```
monkbroc, line 829: You could just do `includePaths.forEach((file) => files.add(file));` instead of putting them in another temporary array.

src/cmd/cloud.js
```diff
@@ -779,11 +779,24 @@ module.exports = class CloudCommand extends CLICommandBase {
...
+		files.forEach((file) => {
+			// source relative to the base directory of the fileMapping (current directory)
+			const source = path.relative(fileMapping.basePath, file);
+			const target = path.relative(dirname, file);
+			fileMapping.map[target] = source;
+		});
+	}
+
+	_getDefaultIncludes(files, dirname, { followSymlinks }) {
 		// Recursively find source files
```
eberseth, line 800: Technically speaking, the default list should probably only have the following specific patterns that are found in src/ and lib/src only:

- project.properties
- src/*.[h|c](pp|xx)?
- src/*.ino
- src/build.mk
- lib/src/*.[h|c](pp|xx)?

We don't want to pick up anything in lib/usage/*.ino for example

src/cmd/cloud.js
```diff
@@ -779,11 +779,24 @@ module.exports = class CloudCommand extends CLICommandBase {
...
-		const ignoreFile = path.join(dirname, settings.dirExcludeFilename);
-		let hasIncludeFile = false;
+		this._getDefaultIncludes(files, dirname, { followSymlinks });
+		this._getCustomIncludes(files, dirname, { followSymlinks });
+		this._getDefaultIgnores(files, dirname, { followSymlinks });
+		this._getCustomIgnores(files, dirname, { followSymlinks });
 
+		// Add files to fileMapping
+		Array.from(files.values()).sort();
+		files.forEach((file) => {
```
monkbroc, line 791: Interestingly this works but you're not getting the benefits of the sort
```suggestion
		const sortedFiles = Array.from(files.values()).sort();
		sortedFiles.forEach((file) => {
```

monkbroc (approved):
Nice work! I made a few suggestions. You can merge and release after addressing them.

src/cmd/cloud.js
```diff
@@ -779,11 +779,24 @@ module.exports = class CloudCommand extends CLICommandBase {
 	 */
 	_processDirIncludes(fileMapping, dirname, { followSymlinks } = {}){
 		dirname = path.resolve(dirname);
+		let files = new Set();
 
-		const includesFile = path.join(dirname, settings.dirIncludeFilename);
-		const ignoreFile = path.join(dirname, settings.dirExcludeFilename);
-		let hasIncludeFile = false;
+		this._getDefaultIncludes(files, dirname, { followSymlinks });
```
eberseth, line 784: Is the intention to stack the settings on top of each other?  For example, the `_getDefaultIncludes()` will add a bunch of patterns then `_getCustomIncludes()` will add to the existing pattern list, and so on?  Or if the custom list is present it will clear any of the default patterns?

eberseth:
This repo and branch is what should be working for Monitor Edge.  There are still files that are not picked up for some reason.
https://github.com/particle-iot-inc/morph-private/tree/feature/cloud-compile

I am using the following command to start the compile:
``` bash
$ npm start -- compile tracker --target 4.1.0 ~/projects/morph/morph-cloudcomp
```

It errored out saying that it was missing lib/memfault/src/memfault-firmware-sdk/components/include/memfault/http/http_client.h
``` bash
src/memfault_demo_http.c:9:10: fatal error: memfault-firmware-sdk/components/include/memfault/http/http_client.h: No such file or directory
    9 | #include "memfault-firmware-sdk/components/include/memfault/http/http_client.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

The other files I see missing for some reason are:

- lib/memfault/examples/memfault_test/src/memfault_particle_user_config.h
- lib/memfault/src/memfault-firmware-sdk/components/include/memfault/core/debug_log.h
- lib/memfault/src/memfault-firmware-sdk/components/include/memfault/core/device_info.h
- lib/memfault/src/memfault-firmware-sdk/components/include/memfault/core/platform/overrides.h
- lib/memfault/src/memfault-firmware-sdk/components/include/memfault/core/platform/reboot_tracking.h
- lib/memfault/src/memfault-firmware-sdk/components/include/memfault/http/http_client.h
- lib/memfault/src/memfault-firmware-sdk/components/include/memfault/http/utils.h
- lib/memfault/src/memfault-firmware-sdk/components/include/memfault/panics/coredump.h

eberseth (approved):

src/cmd/cloud.js
keeramis: The custom list will not clear the list of files added by the default list. The difference is in files added by the user system (_getDefaultIncludes) vs. user (_getCustomIncludes) 

Here's what you can expect from this functionality:
1. `_getDefaultIncludes` will add the default files as provided in the list into a resultant array that will get sent to the cloud compiler. Default list doesn't look at user inputs. It checks the patterns from the system.
2. `_getCustomIncludes` will include more files as per the patterns in the `particle.include` file as specified by the user. Files added in step1 are not cleared.
3. `_getDefaultIgnores` will remove any files as specified by the system (currently only lib/helper/examples/)
4. `_getCustomIgnores` will remove any files as specified by the user in the `particle.ignore` file.

src/cmd/cloud.js
keeramis: @eberseth Thanks for your feedback! I've made the suggested modifications. Just to confirm, are there any other significant patterns we should consider, given that we now focus on `src/*` and `lib/src/*` directories only, excluding other nested directories? With this, we will also no longer support the flat architecture like this - https://github.com/particle-iot/particle-cli/tree/master/test/__fixtures__/projects/legacy-flat

In summary:
1. Flat architecture will not be suppoorted - https://github.com/particle-iot/particle-cli/tree/master/test/__fixtures__/projects/legacy-flat
2. we have this structure of multiple header extensions in one of the tests - should we continue supporting it. https://github.com/particle-iot/particle-cli/tree/master/test/__fixtures__/projects/multiple-header-extensions
3. another one we support is this 'extended' structure where the helper is with in the /src https://github.com/particle-iot/particle-cli/tree/master/test/__fixtures__/projects/extended
4. What about symlinks like this? https://github.com/particle-iot/particle-cli/tree/master/test/__fixtures__/projects/symlink

A simple yes/no from you is sufficient for me to make the expected modifications.

rickkas7 (approved):
I didn't go through the code or run it, but the concept looks good. When is this planned for release so I can update the docs?

keeramis:
@rickkas7 Currently my only ask for feedback is for this - https://github.com/particle-iot/particle-cli/pull/646#discussion_r1226244839 

Once I get clarification on the above, I will get this into the next CLI release.

src/cmd/cloud.js
rickkas7: - No flat format is fine. 
- We never documented all of the other header file extensions but changing the list will likely break libraries in the future; I suspect that list happened when porting Arduino libraries. 
- The helper in src structure seems weird at first, but I suspect that is common for anyone who has divided their project into subdirectories, which is probably not that uncommon. 
- I didn't realize we supported symlinks that way but it could be useful when dealing with shared codebases. But you can also fix this with a build.mk, which might be a better way to do this than symlinks.

keeramis: Cool. Thanks for the inputs. @rickkas7.

keeramis: @eberseth To continue supporting certain legacy project structures, I will keep the changes to the original ones I had. To make changes to remove support for legacy formats, I am leaning towards getting the requirements sorted more clearly and will create a separate PR with those changes. For now, let's keep this PR focused on the particle.include and ignore files.

PR #647, opened 2023-06-22
Fix/add project property for assets

keeramis:
## Description

If a project has a folder named assets, a bundle is inevitably created although it may not be intended by the customer. Therefore, adding a project property called `assetOtaFolder` to the project.properties file such that if this property points to the a folder, then a bundle is created. Otherwise a bundle is not created i.e., assets feature is not enabled.

### A bundle is created after specifying `assetOtaFolder` project property
1. Create a Particle project. 
2. Add a folder called 'assets' which at the root directory of your project
3. Make sure it has the default project.properties file. Add `assetOtaFolder` property to it and point it to the assets folder where you have the 3rd party assets. This folder is preferred to be with in the project root directory
##### project.properties file
```
name=name_of_the_project
assetOtaFolder=name_of_the_folder_which_contains_assets
```
7. Checkout this branch. Run `npm start -- compile argon /path/to/project`
8. Verify that a bundle (.zip file) is created 

### A bundle is not created if `assetOtaFolder` project property is not specified
1. Create a Particle project. 
2. Add a folder named `assets` to the root directory of your project
3. Make sure it has the default project.properties file. Don't add `assetOtaFolder` property to it.
##### project.properties file
```
name=name_of_the_project
```
5. Checkout this branch. Run `npm start -- compile argon /path/to/project`
6. Verify that a bundle (.zip file) is _not_ created

### A bundle is not created if an `assetOtaFolder` project property exists but points to an invalid folder location for assets
1. Create a Particle project. 
2. Add a folder named `assets` to the root directory of your project
3. Make sure it has the default project.properties file. Add `assetOtaFolder` property to it.
##### project.properties file
```
name=name_of_the_project
assetOtaFolder=a_folder_name_that_doesnt_exist
```
7. Checkout this branch. Run `npm start -- compile argon /path/to/project`
8. Verify that a bundle (.zip file) is _not_ created

## Completeness

- [x] User is totes amazing for contributing!
- [x] Contributor has signed [CLA](https://docs.google.com/a/particle.io/forms/d/1_2P-vRKGUFg5bmpcKLHO_qNZWGi5HKYnfrrkd-sbZoA/viewform)
- [x] Problem and solution clearly stated
- [x] Tests have been provided
- [x] Docs have been updated
- [x] CI is passing

PR #650, opened 2023-06-28
Rework particle bundle command

keeramis:
## Description

This PR reworks `particle bundle` command such that it gets the assets from the `--assets /path/to/assets/dir` or the `assetOtaFolder` property of the `--assets /path/to/project.properties`. If `--assets` option is not specified, this command checks if there is a project.properties file in the current working directory.

An example project
```
myProj
    ├── src
    │   ├── app.ino
    └── coverage
    │   ├── ...
    └── assets
    │   ├── asset1.txt
    │   ├── asset2.txt
    │   ├── asset3.txt
    └── target
    │   ├── ...
    └── project.properties
    └── README.txt
```
and its respective `project.properties` file
```
name=myProj
platform=argon
assetOtaFolder=assets
```

Please note that the user has to manually add `assetOtaFolder` property to the project.properties file of their project. The folder path should be relative to the project.properties file of their project, meaning it should be relative to the root of the project.

For more reference
https://www.notion.so/particle/Asset-OTA-CLI-SDD-2483fe9397fa4f30822852b71cad6957

## How to Test

Usage Examples
```
particle bundle /path/to/app.bin --assets /path/to/assets/folder
particle bundle /path/to/app.bin --assets /path/to/project.properties
particle bundle /path/to/app.bin (this will get the assets if there is a project.properties in the current directory and it has a valid assetOtaFolder value)
```

https://github.com/particle-iot/particle-cli/pull/647

## Completeness

- [x] Docs have been updated
- [x] CI is passing


src/cmd/bundle.test.js
```diff
@@ -1,4 +1,6 @@
 const path = require('path');
+const sinon = require('sinon');
+const sandbox = sinon.createSandbox();
```
monkbroc, line 3: Sinon is a sandbox by itself. You can do `sinon.stub` then in the `afterEach` do `sinon.restore`

src/cmd/bundle.js
```diff
@@ -35,13 +35,44 @@ module.exports = class BundleCommands extends CLICommandBase {
...
+			throw new Error('No project.properties file found in the current directory. ' +
+				'Please specify the assets directory using --assets option');
+		}
+		const propFile = await utilities.parsePropertyFile(projectPropertiesPath);
+		if (propFile.assetOtaFolder && propFile.assetOtaFolder !== '') {
+			// get the assets folder relative to the project.properties file
+			const assetsDir = path.join(path.dirname(projectPropertiesPath), propFile.assetOtaFolder);
+			const stats = await fs.stat(assetsDir);
+			if (stats.isDirectory()) {
+				return assetsDir;
```
monkbroc, line 71: No need to check that the folder exists as this is done in `_getAssets`

src/cmd/bundle.js
```diff
@@ -35,13 +35,44 @@ module.exports = class BundleCommands extends CLICommandBase {
...
+			}
+		}
```
monkbroc, line 73: You need to handle the case where there is no asset path in the project.properties. Give a hint to add the assetOtaFolder key to the file.

monkbroc (commented):
Some pre-review comments

src/cmd/bundle.js
keeramis: Wanted to check this was a directory - added the check later.

src/cmd/bundle.test.js
keeramis: Point noted. Actually I forgot to remove them after some testing so removed now.

keeramis:
No strong opinions on the wording, folks. Please suggest if you have more user friendly error messages to show.

src/cmd/bundle.js
```diff
@@ -35,17 +35,52 @@ module.exports = class BundleCommands extends CLICommandBase {
...
+				'Please specify the assets directory using --assets option');
+		}
+		const propFile = await utilities.parsePropertyFile(projectPropertiesPath);
+		if (propFile.assetOtaFolder && propFile.assetOtaFolder !== '') {
+			// get the assets folder relative to the project.properties file
+			return path.join(path.dirname(projectPropertiesPath), propFile.assetOtaFolder);
+		} else if (propFile.assetOtaFolder === '') {
+			throw new Error('assetOtaFolder property in project.properties is empty.');
+		} else if (!propFile.assetOtaFolder) {
+			throw new Error('No assetOtaFolder property found in project.properties.');
```
monkbroc, line 72: No need for 2 conditions. You can remove the empty string one and just keep the falsy condition. A call to action is better than an error message. `Add assetOtaFolder to your project.properties in order to bundle assets`

Make sure to also update the help for `particle bundle` to explain the new syntax: when `--asset` is missing it uses the path from `project.properties`. Add a note to the epilogue about the assetOtaFolder something like: `Add assetOtaFolder=assets to your project.properties to bundle the assets from the assets directory.`

monkbroc (changes requested):
I'll ask you to rename the property for the asset OTA location. I looked in the source code for the CLI and the bundle command is the only place where we talk about "folders". Everywhere else we refer to "directories". So rename the property to `assetOtaDir` and update every place that refers to folder to directory (or dir for short in variable names). Even replace the instance in the cli/library.js help for consistency.

src/lib/utilities.js
```diff
@@ -382,6 +383,12 @@ module.exports = {
 			return new Error(_.isArray(err) ? err.join('\n') : err);
 		}
 		return err;
+	},
+
+	parsePropertyFile(propPath) {
+		const savedProp = fs.readFileSync(propPath, 'utf8');
```
hugomontero, line 389: I'd rather use `await fs.readFile(...)` instead of `readFileSync` the reason is: as far I know `readFileSync` stops the whole eventLoop until it completes the action.

Here's an example of how to use the `readFile` 👉  https://github.com/particle-iot/particle-cli/blob/master/src/cmd/binary.js#L61

keeramis: That way my initial idea but there are other places in the file which are using readFileSync so I kept it consistent with them other than a readFile which waits for the promise to be resolved.

src/cmd/cloud.js
```diff
@@ -690,8 +689,7 @@ module.exports = class CloudCommand extends CLICommandBase {
 		for (const file of files) {
 			const propPath = path.join(file, 'project.properties');
 			try {
-				const savedProp = await fs.readFile(propPath, 'utf8');
-				const savedPropObj = propertiesParser.parse(savedProp);
+				const savedPropObj = await utilities.parsePropertyFile(propPath);
```
hugomontero, line 692: Given that now we have this new property to get assets directory, is this function needs to check if the directory path?
I saw in `_getAssets` function we are doing the same thing and handling in a better way the errors:
https://github.com/particle-iot/particle-cli/pull/650/files#diff-2a5669d3f56c5e923bddc240ef52da8109cb697d4c8707528f098b505b466796R80

I'd say just make either this function returns the path if we have the project property `assetOtaFolder` and let _getAssets do the validations (that now are happening) or remove the validations from _getAssets and put those in here.

Also I saw that in some places we use files[0] and corresponds with the tests that you put. Which is the use case where files will be more than one in this specific function? 

Here's the thing I'm talking about with files[0] 👉  https://github.com/particle-iot/particle-cli/pull/650/files#diff-94014247d62bc2951e9e7135b57b09171b6570579119026b858b7772feb84d98L336

keeramis: @hugomontero I get your point. I changed the code so that the error checking is done in _getAssets and we don't worry about it in the cloud's _checkForAssets. It's basically only checking if it's a directory. Nice optimization suggestion. As far as the files[0], I didn't mess with this and has been there. I am not exactly sure why we expect multiple arguments (i think for multiple source files?) but looks like the expectation is that the first argument in that list is the project directory to the specified by the user.

src/cmd/bundle.js
keeramis: I asked chatgpt for some rewording suggestions for the rest

src/cli/bundle.js
```diff
@@ -3,22 +3,23 @@ module.exports = ({ commandProcessor, root }) => {
...
 		handler: (args) => {
 			const BundleCommands = require('../cmd/bundle');
 			return new BundleCommands().createBundle(args);
 		},
 		examples: {
-			'$0 $command myApp.bin --assets /path/to/assets': 'Creates a bundle of application binary and assets from the /path/to/assets folder',
-			'$0 $command myApp.bin': 'Creates a bundle of application binary and assets from the default /assets folder in the current directory if available',
-			'$0 $command myApp.bin --assets /path/to/assets --saveTo myApp.zip': 'Creates a bundle of application binary and assets from the /path/to/assets folder and saves it to the myApp.zip file',
-			'$0 $command myApp.bin --saveTo myApp.zip': 'Creates a bundle of application binary and assets from the default /assets folder in the current directory if available, and saves the bundle to the myApp.zip file'
+			'$0 $command myApp.bin': 'Creates a bundle of application binary and assets. The assets are obtained from the specified project.properties file',
```
monkbroc, line 17: The assets are obtained from the project.properties in the current directory

src/lib/utilities.js
monkbroc: We have lots of fs sync usage in the CLI. It's best to clean it up when we can but async is viral, so changing one function to async might mean a bunch of changes in other parts of the code.

The consequences are minor in the CLI since it's a single user program so I think that's a chore for another day.

monkbroc (approved):
I'm good with the updated command. The sync stuff isn't a blocker as it's a common bad pattern in the CLI.

hugomontero (approved):
👍

PR #653, opened 2023-08-10
Feature/sc 120959/implement device os version api into particle

hugomontero:
Adds the library that helps to download binaries using Device Os Versions API.

### How to test
This new function receives: `platformId`, `version`, an `api` instance and an `ui` instance.

This is an example on how to use
```
const UI = require('../lib/ui'); // ui from src/lib/ui/index.js
const ParticleApi = require('./api'); // from src/cmd/api
const { downloadDeviceOsVersionBinaries } = require('../lib/device-os-version-util');
const settings = require('../../settings');
const auth = 'XYZ'; // your token

async function download() {
  const particleApi = new ParticleApi(settings.apiUrl, { accessToken: auth });
  const ui = new UI();

  const binaries =  await downloadDeviceOsVersionBinaries({
     platformId: 12,
     api: particleApi,
     ui: ui,
     version: 'latest'
   });
   return binaries;
}

download().then(res => console.log(res)).catch(error => console.log(error))

```

** outcome **
In case the binary already exists it will skip the download process, in other case it will show a progress bar tracking the download progress.
In case the command is being used with a output pipe ( 'npm start -- usb list > output.txt') it will not print the progress but a message indicating the file is being downloaded.

### DIFF
--- a/src/lib/device-os-version-util.js
+++ b/src/lib/device-os-version-util.js
@@ -3,49 +3,106 @@ const deviceConstants = require('@particle/device-constants');
 const path = require('path');
 const request = require('request');
 const fs = require('fs-extra');
+const { HalModuleParser } = require('binary-version-reader');
 
 /**
  * Download a file from the given url to the given directory
  * @param url - the url to download from
  * @param directory - the directory to download to
  * @param filename - the filename to use
+ * @param progressBar - an instance of ProgressBar to use for progress updates
  * @returns {Promise<unknown>} - a promise that resolves when the file is downloaded
  */
-function downloadFile(url, directory, filename) {
-	return new Promise(async (resolve, reject) => {
+async function downloadFile(url, directory, filename, progressBar) {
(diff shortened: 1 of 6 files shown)


src/lib/device-os-version-util.js
```diff
@@ -0,0 +1,161 @@
...
+				const downloadedBinaryName = await downloadBinary({
+					platformName: platform.name,
+					module,
+					baseUrl: deviceOsVersion.base_url,
+					version: deviceOsVersion.version,
+					progressBar,
+				});
+				downloadedBinaries.push(path.join(binaryPath, downloadedBinaryName));
+			} else {
+				ui.write(`Skipping download of ${module.filename} as it is already downloaded`);
```
monkbroc, line 145: No need for a message here as this will be the usual case after the first download.

monkbroc (approved):
Looks good. Let's merge it!

PR #654, opened 2023-08-10
Feature/sc 121026/get device info for local flash

monkbroc:
Story details: https://app.shortcut.com/particle/story/121026

Wire up a method to go through the list of devices, return the only one or prompt the user to select one, then return info about the device so we can device what and how to flash it.


src/cmd/usb-util.js
```diff
@@ -1,5 +1,6 @@
 const { getDevice, isDeviceId } = require('./device-util');
 const { systemSupportsUdev, promptAndInstallUdevRules } = require('./udev');
+const ui = require('../lib/ui');
```
hugomontero, line 3: This `ui` is a class so has to be an instance in order to use the methods
Something like this:

```
const UI = require('../lib/ui');
const ui = new UI();
```

It would be good if we inject ui in the functions. In that way we can get the use from `cmd` classes.

src/cmd/flash.test.js
```diff
@@ -0,0 +1,36 @@
...
+			device = {
+				id: '3c0021000947343432313031',
+				platformId: PlatformId.PHOTON,
+				version: '3.3.1',
+				isInDfuMode: false
+			};
+			// sinon.stub(usbUtil, 'getOneUsbDevice').resolves(device);
+		});
+
+		it('returns information about the device', async () => {
```
hugomontero, line 23: I wasn't able to stub this function by uncommenting the line 20. Every time I test it I need to connect a device and it will fail also given that the expected info doesn't match with my device

hugomontero: Another thing regarding this test is even if pass the device connection still open and the test never ends.
image

src/cmd/usb-util.js
hugomontero: A second thought over here is: we can make that `src/lib/ui` returns as default something like `new UI()` in that way there gonna be just one instance in case we need it in different files on the same execution flow hehe.

src/cmd/flash.test.js
monkbroc: Thanks. I rushed pushing the review button before my next meeting. I'll take another pass at these changes.

monkbroc:
I'll close this PR and rebase it on top of your PR

PR #658, opened 2023-08-17
Feature/sc 121029/flash files locally

monkbroc:
This is a first implementation of Flash that uses naive flash steps, assuming the binaries are already in the right which isn't always the case. Hugo, when you have your method to generate flash steps, we can delete my temporary version.


monkbroc:
I see the tests are failing but I have to head out for the evening. I'll keep working on integrating all this together tomorrow.

PR #661, opened 2023-08-21
Local flash

monkbroc:
Add the `particle flash --local` command.

## How to Test

1. `particle flash --local`
    - Cloud compiles the application in the current directory against the default Device OS version for the connected device
    - Flashes it locally
    - Updates Device OS if necessary
2. `particle flash --local --target 4.1.0`
    - Cloud compiles the application in the current directory against the target Device OS
    - Flashes it locally
    - Force an update to Device OS (might lead to a downgrade)
3. `particle flash --local /path/to/project`
    - Same as 1 but specify the project path instead of using the current directory
4. `particle flash --local /path/to/project --target 4.1.0`
    - Same as 2 but specify the project path instead of using the current directory
5. `particle flash --local /path/to/application`
    - Flashes the application binary or bundle of application and assets, or Device OS binary
    - Updates Device OS if necessary
6. `particle flash --local /path/to/src.cpp`
    - Compile sources in the cloud
    - Flash it locally
    - Update Device OS if needed
7. `particle flash --local /path/to/src.cpp --target 4.1.0 --application-only`
    - Compile sources in the cloud against the target Device OS version
    - Don’t flash Device OS, regardless if it’s outdated
8. `particle flash --local tinker`
    - Flash tinker locally
    - Won’t update Device OS since tinker is compiled against a very old Device OS version
9. `particle flash --local tinker --target 4.1.0`
    - Flash tinker locally
    - Force an update to Device OS (might lead to a downgrade)


hugomontero (approved):
👍

PR #666, opened 2023-09-07
Add known issues to e2e readme

CarlosHdez:
After trying the E2E tests, I came with some small updates that could proof helpful for whoever tries them in the future

## Completeness

- [x] Docs have been updated
- [x] CI is passing


wraithan (approved):
Thanks for the update, would be nice to add a command to this repo that does a preflight test for the tests, or maybe one that paves the devices/account/etc so it starts with a known state, but until then, admitting we have a problem and recording it is a good step.

CarlosHdez:
> 
Usually each test suite (.e2e file) has before/after hooks that set the things as expected for their test, but many times while testing as I saw it failing, I would kill the process before they ran, so trying to run them again failed faster but it was my fault

monkbroc (approved):

PR #667, opened 2023-09-11
Invalidate the 128K binary slot on Gen 3 devices when flashing an app compiled against OS >3.1

keeramis:
## Description

Invalidates 128K binary slot by writing 0xFF to the first flash page which invalidates the header and hence does not run the 128K user part. 

## How to Test
1. Point the device-constants to this PR https://github.com/particle-iot-inc/device-constants/pull/36
1. Create a Particle project which logs a certain text such as 'A' as a unique and simple way to identify the user app running.
Example app:
```
#include "Particle.h"
SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);
SerialLogHandler lg(LOG_LEVEL_ALL);

void setup() {
}

void loop() {
    delay(1s);
    Log.info("A");
}
```

3. Flash a Gen3 nRF device such as an Argon with 5.5.0 by running the CLI local flashing command in terminal `particle flash --local /path/to/project-A --target 5.5.0` or using WB. Call this `app_550.bin`
4. Check that the app logs 'A' in the serial console by running `particle serial monitor --follow`. This is the 256k binary running on the device.
5. Compile and flash another app which displays something else such as `B` against 3.0.0 (or a device-os version < 3.1.0). This will populate the 128k binary slot.
6. Reset the device
7. You will see that 'B' will be printed on the console. This is the 128k binary running on the device
8. Use this branch and run `npm start -- flash --local /path/to/app_550.bin --application-only` 
9. You will see that 'A' will be printed on the console. This is the 256k binary running on the device as expected.

https://app.shortcut.com/particle/story/121947/invalidate-the-128k-binary-slot-on-gen-3-devices-when-flashing-an-app-compiled-against-os-3-1?vc_group_by=day&ct_workflow=all&cf_workflow=500127125

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/lib/flash-helper.js
+++ b/src/lib/flash-helper.js
@@ -157,6 +157,7 @@ async function createFlashSteps({ modules, isInDfuMode, platformId }) {
 	const platform = PLATFORMS.find(p => p.id === platformId);
 	const sortedModules = await sortBinariesByDependency(modules);
 	const assetModules = [], normalModules = [], dfuModules = [];
+	let deviceOsVersion = 0;
 	sortedModules.forEach(module => {
 		const data = module.prefixInfo.moduleFlags === ModuleInfo.Flags.DROP_MODULE_INFO ? module.fileBuffer.slice(module.prefixInfo.prefixSize) : module.fileBuffer;
 		const flashStep = {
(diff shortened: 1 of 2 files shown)


keeramis:
The tests pass once device-constants is merged and released

rickkas7 (approved):
Looks good to me. I didn't run it, but the logic looks correct to me.

src/lib/flash-helper.js
```diff
@@ -174,11 +175,31 @@ async function createFlashSteps({ modules, isInDfuMode, platformId }) {
...
+			prefixInfo : {
+				moduleStartAddy : platform.dfu.segments.formerUserPart.address
+			}
+		};
+		const formerUserPartflashStep = {
+			name: 'invalidate-128k-user-part',
+			moduleInfo,
+			data: Buffer.alloc(platform.dfu.segments.formerUserPart.size, 0xFF)
+		};
+		formerUserPartflashStep.flashMode = 'dfu';
```
monkbroc, line 199: You can include `flashMode: 'dfu'` in the object initialization instead of a separate line in this case

src/lib/flash-helper.js
```diff
@@ -174,11 +175,31 @@ async function createFlashSteps({ modules, isInDfuMode, platformId }) {
 			flashStep.flashMode = 'normal';
 			normalModules.push(flashStep);
 		} else {
+			if (moduleType === 'userPart') {
+				deviceOsVersion = module.prefixInfo.depModuleVersion;
+			}
```
monkbroc, line 180: Move the block of code below to inside the `if` here. It's very important that you push the invalidate step to `dfuModules` BEFORE you push the flash application step otherwise it might erase part of the application you just flashed.

src/lib/flash-helper.test.js
```diff
@@ -217,6 +267,24 @@ describe('flash-helper', () => {
...
+			});
+
+			const expected = [
+				preBootloaderStep,
+				bootloaderStep,
+				systemPart1Step,
+				systemPart2Step,
+				userPart1Step,
+				userPartInvalidationStep
+			];
```
monkbroc, line 284: The expected order is
```suggestion
			const expected = [
				preBootloaderStep,
				bootloaderStep,
				systemPart1Step,
				systemPart2Step,
				userPartInvalidationStep,
				userPart1Step
			];
```

monkbroc (changes requested):
Looks good with the comment that the order needs to be changed

monkbroc (approved):
Nice!

PR #670, opened 2023-09-14
Keep flashing if Device OS version for app cannot be determined

monkbroc:
I got `Invalid version. Must be a string. Got type "object".` when the API didn't return a value for the Device OS version for an app. In that case, bypass downloading the Device OS files. This could happen for a development release or because the API call fails (likely what happened to Eric).


keeramis (approved):
Nice find!

PR #671, opened 2023-09-21
Flash Device OS only when version is known to be outdated

monkbroc:
Change based on discussion in Slack: when doing `particle flash --local` or Workbench: Flash application (local), don't update Device OS if we can't read the version. This is most common when the device is in DFU mode. This avoids needlessfly flashing Device OS over and over.


keeramis (approved):
🎉

PR #672, opened 2023-09-25
bug/sc-122390/issue-when-trying-to-upgrade-the-device-os

hugomontero:
## Description
This PR will: 
* Change the moment we open the device in order to flash it (for every step). 
* Add a retry section for control-request, given that sometimes the device is not ready to receive a request
* Add a validation step in order to know if the device is ready to resolve requests.

## How to Test
In order to make a secure test, you should debug the command `flash --local --target` 
* Put a breakpoint over here 👉  https://github.com/particle-iot/particle-cli/pull/672/files#diff-e8709067fea4097603e1ff058dcc7ebe3104ddc5691236b8b2307bcced4724f0R50
* Start debugging the command and right after the debug reach the break point, reset manually your device by pushing the `reset` button (also you can disconnect and reconnect your usb cable)

Outcome:
* Check that the device will try to run the command and if fail will try to reopen the device and try again.

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/src/cmd/usb-util.js
+++ b/src/cmd/usb-util.js
@@ -239,10 +239,6 @@ async function reopenDevice(device) {
 			device = await openDeviceById(id);
 			// check if we can communicate with the device
 			if (device.isOpen) {
-				// attempt to make a request to the device to check if it responds
-				if (!device.isInDfuMode) {
-					await device.getDeviceMode({ timeout: 1500 });
-				}
 				return device;
 			}
 
--- a/src/cmd/usb-util.js
+++ b/src/cmd/usb-util.js
@@ -227,6 +227,32 @@ async function reopenInNormalMode(device, { reset } = {}) {
 	throw new Error('Unable to reconnect to the device. Try again or run particle update to repair the device');
 }
 
+async function reopenDevice(device) {
+	const { id } = device;
+	if (device.isOpen) {
+		await device.close();
+	}
+	const start = Date.now();
+	while (Date.now() - start < REOPEN_TIMEOUT) {
+		await delay(REOPEN_DELAY);
+		try {
+			device = await openDeviceById(id);
+			// check if we can communicate with the device
+			if (device.isOpen) {
+				// attempt to make a request to the device to check if it responds
+				if (!device.isInDfuMode) {
+					await device.getDeviceMode({ timeout: 1500 });
+				}
+				return device;
+			}
+
+		} catch (err) {
+			// ignore error
+		}
+	}
+	throw new Error('Unable to reconnect to the device. Try again or run particle update to repair the device');
+}
+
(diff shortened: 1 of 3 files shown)


src/lib/flash-helper.js
```diff
@@ -38,29 +38,55 @@ async function flashFiles({ device, flashSteps, resetAfterFlash = true, ui }) {
...
 	if (progress) {
 		progress({ event: 'flash-file', filename: name });
 	}
-	await device.updateFirmware(data, { progress, timeout: FLASH_TIMEOUT });
 
-	// wait for the device to apply the firmware
-	await delay(FLASH_APPLY_DELAY);
-	device = await usbUtils.reopenInNormalMode(device, { reset: false });
+	const start = Date.now();
+	while (Date.now() - start < FLASH_TIMEOUT) {
```
hugomontero, line 47: This is the retry section. I'm not sure on doing it by using an artificial timeout or maybe should be better use a retries number.

src/lib/flash-helper.js
```diff
@@ -38,29 +38,55 @@ async function flashFiles({ device, flashSteps, resetAfterFlash = true, ui }) {
...
+			if (device.isInDfuMode) {
+				// put device in normal mode
+				if (progress) {
+					progress({ event: 'switch-mode', mode: 'normal' });
+				}
+				device = await usbUtils.reopenInNormalMode(device, { reset: true });
+			}
+			try {
+				await device.enterListeningMode();
+				await delay(1000); // Just in case
```
hugomontero, line 76: check this 👉  https://github.com/particle-iot/device-os-flash-util/blob/master/lib/usb.js#L64

src/lib/flash-helper.js
hugomontero: In both: `device-os-flash-util` and `web-setup` we use retry number

monkbroc (approved):
Looks good. I'm testing it on my devices.

src/cmd/usb-util.js
```diff
@@ -239,10 +239,6 @@ async function reopenDevice(device) {
 			device = await openDeviceById(id);
 			// check if we can communicate with the device
 			if (device.isOpen) {
-				// attempt to make a request to the device to check if it responds
-				if (!device.isInDfuMode) {
-					await device.getDeviceMode({ timeout: 1500 });
```
hugomontero: In order to test this behavior: 
* Connect an electron without antenna and wait some seconds.
* Attempt to flash the device

** Behavior ** 
The device will stuck trying to get the device Mode and throw errors regarding `Memory allocation error`
The reason is the device is trying to connect to the cloud and can't handle this specific request (but can handle an listeningMode or dfuMode without issues)

PR #675, opened 2023-09-28
Clean up keys commands

keeramis:
## Description

Remove usage of dfu device specs and reduce the complexity of the keys commands by supporting only one protocol per device. Also remove the need to specify the device id for `particle keys` commands if only one device is connected.

## How to Test

1. Run any `particle keys xxx` commands and `particle doctor` command on your device(s) and verify that they are working as expected. 
1. For example, To run `particle keys xxx`, run `npm start -- keys xxx` from project's root directory

https://app.shortcut.com/particle/story/122385/remove-usage-of-dfu-util-in-particle-cli

## Completeness

- [x] Docs have been updated
- [x] CI is passing


src/cmd/keys.js
```diff
@@ -617,27 +588,54 @@ module.exports = class KeysCommand {
...
 		} catch (err) {
 			throw new VError(ensureError(err), 'Unable to get DFU device');
 		}
 	}
 
-	_validateSegmentSpecs(segmentName) {
-		let specs = deviceSpecs[this.dfuId] || {};
-		let params = specs[segmentName];
+	_validateSegments(segmentName) {
+		let segments = this._getAllSegments() || {};
```
hugomontero, line 599: Given that you don't change `segments` and `params` after creation you can put `const` instead of `let`

keeramis: yes! surprised that lint didn't warn this time

hugomontero: Regarding this https://github.com/particle-iot/particle-cli/pull/675/files#r1340443296

You can do:
```suggestion
		const segments = this._getAllSegments();
		const params = segments ? segments[segmentName] : null;
```

hugomontero: Checking more deep the way getAllSegments works you always will receive an object. there is no way to get a null. So the `|| {}` can be removed

src/cmd/flash.js
```diff
@@ -55,29 +55,46 @@ module.exports = class FlashCommand extends CLICommandBase {
...
-		const resetAfterFlash = !factory && modulesToFlash[0].prefixInfo.moduleFunction === ModuleInfo.FunctionType.USER_PART;
-		await flashFiles({ device, flashSteps, resetAfterFlash, ui: this.ui });
+			this.ui.write(`Flashing ${platformName} device ${device.id}`);
+			const resetAfterFlash = !factory && modulesToFlash[0].prefixInfo.moduleFunction === ModuleInfo.FunctionType.USER_PART;
+			await flashFiles({ device, flashSteps, resetAfterFlash, ui: this.ui });
+		} catch (error) {
+			if (device) {
+				await device.close();
+			}
+		}
```
keeramis, line 97: @hugomontero and I figured that device doesn't close and it leaves the terminal hanging if the logic fails in any of the above lines such as `const modulesToFlash = await parseModulesToFlash({ files });` . We are if we should add this try/catch block here in this file for other methods (which is `flashLocal`) or if there is a better way to handle it.

monkbroc (approved):

PR #676, opened 2023-10-02
Asset bundling should be supported only for targeted platforms and Device-OS versions

keeramis:
Story details: https://app.shortcut.com/particle/story/121840


keeramis:
Updated the `app.bin` files in test fixtures to 5.5.0 and above

PR #680, opened 2023-10-12
Improve usb device selector output

keeramis:
## Description

Device selector will now show more details

With user creds and api connectivity available
```

? Which device would you like to select? (Use arrow keys)
❯ seal_penguin [0a10aced202194944a02c4d4] (Photon 2 / P2, LISTENING) 
  morphing_banjo [e00fce6819ef5f971ea9563a] (Argon) 

```

Without api connectivity, it does not show the name
```
? Which device would you like to select? (Use arrow keys)
❯  [0a10aced202194944a02c4d4] (Photon 2 / P2, LISTENING) 
   [e00fce6819ef5f971ea9563a] (Argon) 
```

## How to Test

1. Checkout his branch to test
2. Connect at least two devices
3. Run `npm start -- flash --usb app.bin`
4. Ensure the output is as expected

 NA

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/usb-util.js
+++ b/src/cmd/usb-util.js
@@ -32,7 +32,7 @@ async function _getDeviceInfo(device) {
 		}
 	} catch (err) {
 		if (err instanceof TimeoutError) {
-			return { id, mode: 'UNKNOWN'};
+			return { id, mode: 'UNKNOWN' };
 		} else {
 			throw new Error(`Unable to get device mode: ${err.message}`);
 		}
@@ -198,7 +198,7 @@ async function getOneUsbDevice({ idOrName, api, auth, ui }) {
 					const { id, mode } = await _getDeviceInfo(d);
 					const name = await _getDeviceName({ id, api, auth, ui });
 					return {
-						name: `${name || '<no name>'} [${id}] (${platformForId(d._info.id).displayName}${mode ? ', ' + mode : ''})`,
+						name: `${name || '<no name>'} [${id}] (${platformForId(d._info.id).displayName}${mode ? ', ' + mode : '' })`,
 						value: d
 					};
 				}));
--- a/src/cmd/usb-util.js
+++ b/src/cmd/usb-util.js
@@ -15,6 +15,22 @@ const REOPEN_TIMEOUT = 60000;
 // When reopening a device that was about to reset, give it some time to boot into the firmware
 const REOPEN_DELAY = 500;
 
+
+async function _getDeviceId(device) {
+	let id = null;
+	try {
+		await device.open();
+		return device._id;
+	} catch (err) {
+		// ignore error
+	} finally {
+		if (device.isOpen) {
+			await device.close();
+		}
+	}
+	return id;
+}
+
 /**
  * USB permissions error.
  */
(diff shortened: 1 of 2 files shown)


src/cmd/usb-util.js
```diff
@@ -153,12 +203,15 @@ async function getOneUsbDevice({ idOrName, api, auth, ui }) {
 			name: 'device',
 			message: 'Which device would you like to select?',
 			choices() {
-				return usbDevices.map((d) => {
+				return Promise.all(usbDevices.map(async (d) => {
+					const id = await _getDeviceId(d);
+					const mode = await _getDeviceMode(d);
```
hugomontero, line 208: It might be good if we open the device once and get the mode and the id in one step?
Also that will allow us to have a `parallel` behavior here:

```
[{id, mode}, name] = Promise.all[_getDeviceInfo(d), _getDeviceName({...})]

```

hugomontero (approved):
It Worked great!! I Just put a minor comment

src/cmd/usb-util.js
monkbroc: Good suggestion

src/cmd/usb-util.js
```diff
@@ -15,6 +16,55 @@ const REOPEN_TIMEOUT = 60000;
...
+		}
+	}
+}
+
+async function _getDeviceName({ id, api, auth, ui }) {
+	try {
+		const device = await getDevice({ id, api, auth, ui });
+		return device.name;
+	} catch (err) {
+		// ignore error
```
monkbroc, line 64: Can you see if you can distinguish the 2 cases in this error handler and return the following names:
`<no name>`: the device hasn't been named yet
`<unknown>`: if the API request failed (errored with 404 Not Found because you don't have access to the device or if the API request couldn't be performed because the computer is offline and the device attributes are not in cache)

`particle usb list` does show `<no name>`
```
$ particle usb list
<no name> [18002b000247353138383138] (Photon, LISTENING)
extra_p2 [0a10aced202194944a02c5f4] (Photon 2 / P2)
```

monkbroc (approved):
Looks good. I made a suggestion to handle unknown and no name

src/cmd/usb-util.js
keeramis: yes - let me try that

keeramis: Hope this looks correct - added in the change
If connection is lost
```
? Which device would you like to select? (Use arrow keys)
❯ <unknown> [0a10aced202194944a02c4d4] (Photon 2 / P2, DFU) 
  <unknown> [e00fce6819ef5f971ea9563a] (Argon) 
```
No name
```
? Which device would you like to select? (Use arrow keys)
❯ seal_penguin [0a10aced202194944a02c4d4] (Photon 2 / P2, DFU) 
  <no name> [e00fce6819ef5f971ea9563a] (Argon) 
```

src/cmd/usb-util.js
keeramis: @hugomontero Added a `_getDeviceInfo` instead of two different methods. But to write everything in a single Promise.all() , the `_getDeviceName()` requires `id` which is available from the first call. Let me know if I am missing something.

hugomontero: Is like you said hehe. 👍

PR #691, opened 2023-11-23
Feature/sc 123395/implement create command

hugomontero:
## Completeness

- [ ] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/src/cmd/logic-function.js
+++ b/src/cmd/logic-function.js
@@ -1,9 +1,13 @@
+const os = require('os');
+const path = require('path');
 const ParticleAPI = require('./api');
+const CLICommandBase = require('./base');
 const VError = require('verror');
 const settings = require('../../settings');
 const { normalizedApiError } = require('../lib/api-client');
-const CLICommandBase = require('./base');
-const os = require('os');
+const { copyAndReplaceTemplate, hasTemplateFiles } = require('../lib/template-processor');
+
+const logicFunctionTemplatePath = __dirname + '/../../assets/logicFunction';
 
 /**
  * Commands for managing encryption keys.
@@ -19,7 +23,7 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
 		try {
 			const logicFunctions = await api.getLogicFunctionList({ org });
 			const orgName = getOrgName(org);
-			const list = logicFunctions['logic_functions'];
+			const list = logicFunctions.logic_functions;
 			if (list.length === 0) {
 				this.ui.stdout.write(`No Logic Functions currently deployed in your ${orgName}.`);
 				this.ui.stdout.write(`To create a Logic Function, see \`particle logic-function create\`.${os.EOL}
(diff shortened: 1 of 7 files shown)


src/cmd/logic-function.js
```diff
@@ -46,9 +52,129 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
...
+			const { overwrite } =  await this.ui.prompt([question]);
+			if (!overwrite) {
+				this.ui.stdout.write(`Aborted.${os.EOL}`);
+				process.exit(0);
+			}
+		}
+	}
+
+	/** Recursively copy and replace template files */
+	async _copyAndReplaceLogicFunction({ logicFunctionName, logicFunctionSlugName, description, templatePath, destinationPath }){
```
hugomontero, line 142: This can be refactored. I'll do that in further branches

src/cmd/logic-function.js
```diff
@@ -46,9 +52,129 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
...
+		this.ui.stdout.write(`- \`particle logic execute\` to run your Function${os.EOL}`);
+		this.ui.stdout.write(`- \`particle logic deploy\` to deploy your new changes${os.EOL}`);
+		return createdFiles;
+	}
+
+	async _validateExistingName({ api, org, name }) {
+		// TODO (hmontero): request for a getLogicFunctionByName() method in the API
+		let existingLogicFunction;
+		try {
+			const logicFunctionsResponse = await api.getLogicFunctionList({ org });
```
keeramis, line 111: IMO, This can directly use the `list` method from this class.

Until we may have an API to get the name and id separately, I added [this internal function](https://github.com/particle-iot/particle-cli/compare/feature/sc-123396/implement-execute-command-with-eventdata...feature/logic-function-get#diff-14f13d7b42c6c8b6494d7858d3a4191c54c8a528a3689709dd0ac6c2abbf92a7R96-R115) to fetch the full list of the logic functions, and obtain the `name` if `id` is given and vice versa. @hugomontero raised a concern about downloading a potentially _large_ list of logic functions and how that could impact the behavior or performance.

I made the above changes in the future branch: https://github.com/particle-iot/particle-cli/tree/feature/logic-function-get

src/cmd/logic-function.js
```diff
@@ -46,9 +52,129 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
...
+		name = name.trim();
+		// ask for description
+		const question = {
+			type: 'input',
+			name: 'description',
+			message: 'Add a description for your Function (optional)'
+		};
+		const result =  await this.ui.prompt([question]);
+		const description = result.description;
+		const slugName = name.toLowerCase().replace(/\s/g, '-');
```
monkbroc, line 82: This is the slugify function from api-service. Can you use that? Copy it to a helper
```
	/**
	 * Converts a string to the slug version by replacing
	 * spaces and underscores with dashes, changing
	 * to lowercase, and removing anything other than
	 * numbers, letters, and dashes
	 * @param  {String} str string to slugify
	 * @return {String}     slugified version of str
	 */
	slugify(str) {
		const slug = str.trim().toLowerCase()
			// replace every group of spaces and underscores with a single hyphen
			.replace(/[ _]+/g, '-')
			// delete everything other than letters, numbers and hyphens
			.replace(/[^a-z0-9-]/g, '');
		return slug;
	},
```

hugomontero: Done here! 👉 https://github.com/particle-iot/particle-cli/pull/691/commits/348a5832b49b97b5bd6d5a9c33610b6d1d9b23b8

PR #696, opened 2023-12-04
Feature/logic function get

keeramis:
## Description

This PR points to the execute command PR .

This PR does the following:
- Implements the `get` command which downloads a logic function. The logic function can be specified using the `--name` or `--id` parameter on the CLI. Or it can be selected from the list of available logic functions
- Uses `list` to get the `name` and `id` mapping to obtain one from the other
- Improves color representation on the UI messages
- Add a boolean parameter called `display` to the `list` method which controls whether to return the list for other helper methods or display it to the UI for the user
- Add hyphen similar to `create` command if there are spaces (This will change as the specs change)
- TODO: Remove the organization option in `help` (compare with the other PR)

## How to Test

1. Do the setup drill (checkout the branch, install dependencies, etc)
2. Run this command `npm start -- logic-function get` and tinker with it using options `--name`, `--id`, `--org`
3. Usage includes but not limited to the following
    ```
    npm start -- logic-function get
    npm start -- logic-function get --name myLF
    npm start -- logic-function get --id xxxx-xxxx-xxxx-xxxx-xxxx
    npm start -- logic-function get --org Particle
    ```

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing


src/cmd/logic-function.js
```diff
@@ -48,97 +56,224 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
...
-
 	async get({ org, name, id }) {
-		// 1. Get the list of logic functions to download from
-		//const list = await this.list();
-		console.log(org, name, id);
-		// 2. Select one using picker
-		// 3. Download it to files. Take care of formatting
+		const api = createAPI();
+		({ name, id } = await this._getLogicFunctionIdAndName(org, name, id));
+		const slugName = name.toLowerCase().replace(/\s/g, '-');
```
hugomontero, line 62: let's change it with `slugify` from `utils`

src/cmd/logic-function.test.js
```diff
@@ -245,8 +242,417 @@ describe('LogicFunctionCommands', () => {
...
+				.intercept('/logic/functions', 'GET')
+				.reply(200, { logic_functions: logicFunctions });
+
+			const stubGet = nock('https://api.particle.io/v1',)
+				.intercept('/logic/functions/0021e8f4-64ee-416d-83f3-898aa909fb1b', 'GET')
+				.reply(200, { logic_function: logicFunc1.logic_functions[0] });
+
+			await logicFunctionCommands.get({ name: 'LF1' });
+
+			expect(fs.existsSync(path.join(process.cwd(), 'LF1', 'LF1.js'))).to.be.true;
```
hugomontero, line 423: Pretty sure that's making linux tests fail.
I'd rather use `fixtures/__root__/...` folder for creation. Let's try to repro the test error in a linux machine.
the only thing that came to my mind now is `lf1` dir and files should be in lower case since we're converting the name with `slugify` function.

src/cmd/logic-function.test.js
```diff
@@ -245,8 +242,419 @@ describe('LogicFunctionCommands', () => {
...
+			logicFunctions.push(logicFunc2.logic_functions[0]);
+
+			const res = await logicFunctionCommands._getNameFromId('0021e8f4-64ee-416d-83f3-898aa909fb1c', logicFunctions);
+
+			expect(res).to.equal(null);
+		});
+	});
+
+	describe('get', async () => {
+		beforeEach(() => {
```
hugomontero, line 358: use `afterEach` to return everything to normal after your test

src/cmd/logic-function.js
```diff
@@ -20,25 +20,34 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
 		super(...args);
 	}
 
-	async list({ org }) {
+	async list({ org, display = true }) {
```
hugomontero, line 23: This sounds like we want to have a `display` argument in the command.
I think this is not a good approach for that (we're mixing responsibilities over here)
Put a TODO to fix it by creating a list function that returns the thing we want to return in this case I think the goal is return a `choice`  object.

src/cmd/logic-function.js
```diff
@@ -47,97 +56,227 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
...
-		// 2. Select one using picker
-		// 3. Download it to files. Take care of formatting
+		const api = createAPI();
+		({ name, id } = await this._getLogicFunctionIdAndName(org, name, id));
+		const slugName = slugify(name);
+
+		try {
+			const logicFunction = await api.getLogicFunction({ org, id });
+
+			// TODO: Address spaces in the name for Windows
```
hugomontero, line 67: this todo is not required since the `slugify`

hugomontero (approved):
I'd say let's merge it since we agree to change the dir creation in other story as well the `list` function

monkbroc (commented):
Keerthy and I reviewed this PR over Zoom. Keerthy has some notes on things to improve.

src/cmd/logic-function.js
```diff
@@ -285,6 +359,103 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
...
+			throw new Error('No logic functions found');
+		}
+		const answer = await this._prompt({
+			type: 'list',
+			name: 'logic_function',
+			message: 'Which logic function would you like to download?',
+			choices : list,
+			nonInteractiveError: 'Provide name for the logic function'
+		});
+		return answer.logic_function;
```
keeramis, line 437: @monkbroc I know we discussed to return name and id here. I didn't make that change to keep the `_getLogicFunctionIdAndName` look consistent in the way it calls the `_getXFromY` methods - Kindly let me know if this should have been different.

lf1/lf1.js
```diff
@@ -0,0 +1,4 @@
+import Particle from 'particle:core';
```
hugomontero, line 1: This and `lf1.logic.json` should be removed or put those in __fixtures__

src/cmd/logic-function.js
```diff
@@ -18,123 +18,196 @@ const logicFunctionTemplatePath = path.join(__dirname, '/../../assets/logicFunct
...
 
+	_printListHelperOutput() {
+		this.ui.stdout.write(`No Logic Functions currently deployed in your ${getOrgName(this.org)}.${os.EOL}`);
+		this.ui.stdout.write(`${os.EOL}`);
+		this.ui.stdout.write('To create a Logic Function, see  ' + this.ui.chalk.yellow('\'particle logic-function create\'.') + `${os.EOL}`);
+		this.ui.stdout.write('To download an existing Logic Function, see ' + this.ui.chalk.yellow('\'particle logic-function get\'.') + `${os.EOL}`);
+	}
+
+	_printListOutput({ logicFunctionsList }) {
+		this.ui.stdout.write(`Logic Functions currently deployed in your ${getOrgName(this.org)}:${os.EOL}`);
```
monkbroc, line 46: General comment for another PR: make `getOrgName` return "your Sandbox" or the name of the org so that the messages print like this:
Logic Functions currently deployed in your Sandbox
Logic Functions currently deployed in Particle

instead of currently which would show
Logic Functions currently deployed in your Particle

monkbroc (approved):
Retest the get and create functions after the latest changes and once they work, go ahead and merge

PR #700, opened 2023-12-13
CLI Logic Function - Enable

keeramis:
## Description

Implements `enable` command to enable a Logic Function

## How to Test

`particle lf enable --name myLF`

https://app.shortcut.com/particle/story/124133/implement-enable-command

## Completeness

- [x] Docs have been updated
- [x] CI is passing


src/cmd/logic-function.js
```diff
@@ -334,19 +334,23 @@ module.exports = class LogicFunctionsCommand extends CLICommandBase {
 		console.log(org, filepath);
 	}
 
-	async disable({ org, name, id }) {
+	async disable({ org, name, id }, status) {
```
hugomontero, line 337: I'd rather to extract the inner functions that `disable` / `enable` the logicFunction instead of giving a new responsibility to `disable` function.

hugomontero: What I mean with that is to create a new function `enable` and extract the logic to change the status.
Something important if the user execute twice the `enable` command the logic-function should remains enable. Same thing with `disable` command

keeramis: >Something important if the user execute twice the enable command the logic-function should remains enable.


It will still remains to be enabled. It's the same functionality other than the bit flip so I think we can keep it under the same function? The only other difference is the output logs which is different for enable and disable. The important point is that the functionality is the same with different bits for false/true for the logic_function.enabled .

keeramis: I have also tested enabling in the console, disabling in the CLI and vice verse.
And enabling and disabling multiple times work as expected

hugomontero: Here the thing is the function name is `disable` and do something else than `disable` a logicFunction. That's the main reason.

hugomontero: if you want to keep the same functionality then change the function name to `updateStatus` or something like specifies the actual thing the function is doing.

hugomontero (approved):
👍

PR #706, opened 2023-12-21
Add end to end tests for CLI Logic commands

monkbroc:
## How to Test

`npm run test:e2e -- --grep "Logic Function Commands"`


test/e2e/logic-function.e2e.js
```diff
@@ -0,0 +1,227 @@
+const { expect } = require('../setup');
```
hugomontero, line 1: In general this test might fail on concurrent executions especially the `list` one

hugomontero (approved):
👍

PR #726, opened 2024-04-15
fix: [sc-127227] Support renamed assets

monkbroc:
Story details: https://app.shortcut.com/particle/story/127227

If an asset is renamed but not otherwise modified, the CLI won't flash the newly named asset to the device, leading to the device being in safe mode.

Steps to test:
- Compile a project with one asset `test.txt`
- `particle flash --local app.zip`
- Rename the asset `test2.txt` and recompile
- `particle flash --local app.zip`
- Check that `test2.txt` is properly flashed.

I made an additional change to avoid resetting the device an additional when flashing assets in normal mode. The 2 scenarios are either that the asset is flashed, then the device reboots on its own, or the asset is skipped so an addtional reset causes the application to restart one more time than necessary.

**Hugo, can you think of a downside to this change?**

### DIFF
--- a/src/lib/flash-helper.js
+++ b/src/lib/flash-helper.js
@@ -329,7 +329,7 @@ async function createFlashSteps({ modules, isInDfuMode, factory, platformId }) {
 
 function _skipAsset(module, existingAssets) {
 	const hashAssetToBeFlashed = _get256Hash(module);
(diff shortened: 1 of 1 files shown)


hugomontero (approved):
Looks great! 👍

PR #728, opened 2024-04-15
Fix the asset check code

monkbroc:
Additional fixes on top of #726
* Compare only the basename of the asset filename
* Move the command to enter listening mode to after checking if an asset should be skipped to avoid putting the device in listening mode when skipping all the assets.

### DIFF
--- a/src/lib/flash-helper.js
+++ b/src/lib/flash-helper.js
@@ -66,6 +66,12 @@ async function _flashDeviceInNormalMode(device, data, { name, progress, checkSki
 				}
 				return device;
 			}
+			try {
+				await device.enterListeningMode();
+				await delay(1000); // Just in case
+			} catch (error) {
+				// ignore
+			}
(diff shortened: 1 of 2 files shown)


src/lib/flash-helper.js
```diff
@@ -66,6 +66,12 @@ async function _flashDeviceInNormalMode(device, data, { name, progress, checkSki
 				}
 				return device;
 			}
+			try {
```
hugomontero, line 69: Makes more sense this approach since prepareToFlash always puts the device in listening for nonDFU modes. 
I was thinking about a final reset after the whole process could be also a good way to have the device back to normal but would be too much since in all the times the device is being resetting when it flashes something hehe.

hugomontero (approved):
👍

src/lib/flash-helper.js
monkbroc: Flashing with control requests resets by itself so it should be in a good state without an extra reset.

PR #736, opened 2024-05-14
Fix wifi configuration on serial

keeramis:
## Description

Run serial wifi configuration via control requests for gen3 and above. Use serial commands for gen2.

## How to Test

1. Setup this branch
2. Connect a single or multiple devices
3. Run the following commands
`npm start -- serial wifi`
`npm start -- serial wifi --file creds.json`
where creds.json looks like
```
{
  "network": "my_ssid",
  "security": "WPA2_AES",
  "password": "my_password"
}

```

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/src/cmd/serial.js
+++ b/src/cmd/serial.js
@@ -21,6 +21,7 @@ const usbUtils = require('./usb-util');
 const { platformForId } = require('../lib/platform');
 const { FirmwareModuleDisplayNames } = require('../lib/require-optional')('particle-usb');
 const WifiControlRequest = require('../lib/wifi-control-request');
+const semver = require('semver');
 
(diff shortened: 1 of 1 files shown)


monkbroc (approved):
Here it makes sense to use generation/version check since we need to ask questions to users so it's harder to do feature detection.

hugomontero:
Control requests prevents the user to pre-configure the wifi if the SSID is not present on the area.
The use case that made us to rollback the first change was:
On production line some costumers requires to pre-configure the wifi credentials de devices will use in the work area.

monkbroc:
Good call Hugo. Keerthy changed the fix to use control requests for DVOS >6 (since serial commands are gone there) and she'll figure out with Scott a better solution.

PR #753, opened 2024-07-22
Automate Protected Devices to run `particle usb` commands

keeramis:
## Description

Support Protected Devices while running `particle usb xxx` commands
All of the `particle usb xxx` commands enable Protection on the device after the command is finished except these two:
- `particle usb reset`
- `particle usb dfu`

Also, added a fix for `particle usb cloud-status device-name` which was throwing an error due to an already existing bug when using `device-name`

## How to Test

Run `particle usb xxx` commands on Protected, Protected (Service Mode), and Open Devices.
Run them when the device is in normal mode and dfu mode. 
- Are you seeing the expected output?
- Are you seeing the device recover if it errors?
- Does it show helpful error messages?

https://github.com/particle-iot/particle-cli/pull/754
https://app.shortcut.com/particle/story/129067/add-generic-helpers-for-a-device-to-put-into-service-mode-and-back-to-perform-a-cli-operation

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing


src/cmd/usb-util.js
```diff
@@ -124,8 +124,11 @@ async function executeWithUsbDevice({ args, func, dfuMode = false } = {}) {
 	} finally {
 		if (deviceIsProtected) {
 			device = await reopenDevice(device);
-			// FIXME: ignore errors
-			await deviceProtectionHelper.turnOffServiceMode(device);
+			try {
+				await deviceProtectionHelper.turnOffServiceMode(device);
+			} catch (error) {
+				// FIXME: ignore errors
```
monkbroc, line 130: You can get rid of the FIXME.
Which use case did you see that needs errors here to be ignored?

src/cmd/usb-util.js
```diff
@@ -124,8 +124,11 @@ async function executeWithUsbDevice({ args, func, dfuMode = false } = {}) {
...
+			}
 		}
```
keeramis, line 132: `executeWithUsbDevice` is called from `forEachUsbDevice`, and this error will be propagated to `forEachUsbDevice` and if not handled properly, it will show an internal error to the user (such as `IN control transfer failed` which is likely useless to the user because they cant act on it)

src/cmd/usb.js
```diff
@@ -186,52 +188,56 @@ module.exports = class UsbCommand extends CLICommandBase {
...
+			if (Date.now() - started >= timeout) {
+				throw new Error('timed-out waiting for status...');
+			}
+		} else {
+			try {
+				status = await device.getCloudConnectionStatus();
+			} catch (error) {
+				// FIXME
+				throw new Error(error);
+			}
```
monkbroc, line 237: This is unnecessary. If you don't want to catch the error, remove the try and catch blocks.

test/e2e/usb.e2e.js
```diff
@@ -179,16 +181,21 @@ describe('USB Commands [@device]', function cliUSBCommands(){
 			expect(exitCode).to.equal(0);
 		});
 
-		it('Lists connected devices when signed-in to a foreign account', async () => {
+		xit('Lists connected devices when signed-in to a foreign account', async () => {
+			await cli.logout();
 			await cli.loginToForeignAcct();
 			const { stdout, stderr, exitCode } = await cli.run(args);
+			console.log('stdout:', stdout);
```
monkbroc, line 188: Remember to remove console logs

test/e2e/usb.e2e.js
```diff
@@ -179,16 +181,21 @@ describe('USB Commands [@device]', function cliUSBCommands(){
 			expect(exitCode).to.equal(0);
 		});
 
-		it('Lists connected devices when signed-in to a foreign account', async () => {
+		xit('Lists connected devices when signed-in to a foreign account', async () => {
```
monkbroc, line 184: Why does this test not work anymore?

test/lib/cli.js
```diff
@@ -182,7 +184,7 @@ module.exports.getCloudConnectionStatus = async () => {
 
 module.exports.waitUntilOnline = async () => {
 	const { run } = module.exports;
-	const args = ['usb', 'cloud-status', DEVICE_ID, '--until', 'connected', '--timeout', 5 * 60 * 1000];
+	const args = ['usb', 'cloud-status', DEVICE_ID, '--until', 'connected', '--timeout', 5000];
```
monkbroc, line 187: Waiting 5 seconds for the device to connect to the cloud during E2E tests isn't enough. Why the change?

monkbroc (commented):
Looks good but I have some questions on the changes to tests

src/cmd/usb-util.js
keeramis: @monkbroc 
`executeWithUsbDevice` is called from `forEachUsbDevice`, and this error will be propagated to `forEachUsbDevice` and if not handled properly, it will show an internal error to the user (such as IN control transfer failed which is likely useless to the user because they cant act on it)

Use case: Enabling Protection on the device immediately after running the safe mode command caused an `IN control transfer failed` error. I added a `waitForDeviceToReboot` after with safe mode command to let it work correctly but I realized that an error (if any) can propagate to `forEachUsbDevice` when using `turnOffServiceMode`

keeramis: Ignoring this error means that an internal error will not be thrown to the user, and device will likely be in Service Mode if any error occurs here that is preventing turning off Service Mode

src/cmd/usb.js
```diff
@@ -186,52 +187,51 @@ module.exports = class UsbCommand extends CLICommandBase {
...
+		try {
+			status = await executeWithUsbDevice({
+				args: { idOrName: device }, // device here is the id
+				func: (dev) => this._cloudStatus(dev, until, timeout),
+			});
+		} catch (error) {
+			this.stopSpin();
+			throw new Error(error.message);
+		} finally {
+			this.stopSpin();
```
hugomontero, line 204: `finally` will always be executed, so you can remove `this.stopSpin()` from catch since you are doing it in the `finally` block

Check here 👉  [try-catch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch)

src/cmd/usb.js
```diff
@@ -260,7 +260,7 @@ module.exports = class UsbCommand extends CLICommandBase {
 		// define output array with logs to prevent interleaving with the spinner
 		let output = [];
 
-		await forEachUsbDevice(args, usbDevice => {
+		await forEachUsbDevice(args, async (usbDevice) => {
```
hugomontero, line 263: The structure here didn't change, why we put the `async` here?

src/cmd/usb.js
```diff
@@ -186,52 +187,51 @@ module.exports = class UsbCommand extends CLICommandBase {
...
-			.catch(error => {
-				if (until){
-					return deviceMgr.close()
-						.then(() => delay(1000))
-						.then(() => this.cloudStatus(args, started));
-				}
-				throw error;
-			})
-			.finally(() => deviceMgr.close());
+		return status;
```
hugomontero, line 234: How about:

```
const started = Date.now();

if (until) {
	while (Date.now() - started < timeout) {
		try {
			let status = await device.getCloudConnectionStatus();
			status = status.toLowerCase();
			if (status === until) {
				return status;
			}
		} catch (error) {
			// ignore error and keep trying until timeout has elapsed
		}
	}
	throw new Error('timed-out waiting for status...');
} else {
	return device.getCloudConnectionStatus();
}
```

keeramis: this works too. I preferred a single point of return

src/cmd/usb.js
keeramis: I wanted to ensure the error thrown does not interfere with the spinner

src/cmd/usb-util.js
```diff
@@ -432,10 +467,15 @@ async function forEachUsbDevice(args, func, { dfuMode = false } = {}){
...
-							args: { idOrName : usbDevice.id },
+							args: { idOrName : usbDevice.id, api: args.api, auth: args.auth },
 							func,
 							dfuMode
 						});
+						// This is particularly useful if the device handle is lost in the executed function
+						// Example: particle usb reset
+						if (usbDevice?.isOpen) {
+							await usbDevice.close();
+						}
```
keeramis, line 478: Still unclear to me why the handle is lost in the executeWithUsbDevice but this was needed to close the hanging handle

test/e2e/usb-protected.e2e.js
```diff
@@ -0,0 +1,440 @@
+const capitalize = require('lodash/capitalize');
```
keeramis, line 1: Comment for this file as a whole. I intentionally included a few redundant checks in the `after` of the previous describe block as well as `before` of the current describe block for increases robustness

src/cmd/usb-util.js
```diff
@@ -123,9 +124,16 @@ async function executeWithUsbDevice({ args, func, dfuMode = false } = {}) {
...
+			device = await reopenDevice({ id: deviceId });
+			if (!device.isInDfuMode) {
+				device = await waitForDeviceToReboot(device.id);
+			}
+			try {
+				// Only works for 6.1.2 and later
+				await deviceProtectionHelper.turnOffServiceMode(device);
+			} catch (error) {
+				// FIXME: ignore errors
+			}
```
keeramis, line 136: I added this to ensure the device is ready for further operations after it finished the respective CLI operation. I forgot which use case needed this change but it was during e2e tests that I found out that this is a good check to have. IMO it adds more robustness. Please let me know if anyone disagrees.
```
if (!device.isInDfuMode) {
        device = await waitForDeviceToReboot(device.id);
}
```

src/cmd/usb-util.js
```diff
@@ -134,6 +137,33 @@ async function executeWithUsbDevice({ args, func, dfuMode = false } = {}) {
...
+			// Check device readiness
+			await device.getDeviceId();
+			return device;
+		} catch (error) {
+			// ignore error
+		}
+	}
+}
+
+async function _delay(ms){
```
hugomontero, line 168: We already have a delay function 👉 https://github.com/particle-iot/particle-cli/blob/0fcc1d661ad7331d20eb9578a798a20cfcfe2f3b/src/lib/utilities.js#L139 Can you use it instead of creating a new one?

test/lib/cli.js
```diff
@@ -52,9 +52,11 @@ module.exports.login = () => {
 	return run(['login', '-u', USERNAME, '-p', PASSWORD], { reject: true });
 };
 
-module.exports.loginToForeignAcct = () => {
+module.exports.loginToForeignAcct = async () => {
 	const { run } = module.exports;
-	return run(['login', '-u', FOREIGN_USERNAME, '-p', FOREIGN_PASSWORD], { reject: true });
+	await run(['login', '-u', FOREIGN_USERNAME, '-p', FOREIGN_PASSWORD], { reject: true });
+	const s = await run(['whoami'], { reject: true });
+	console.log('Logged in as:', s);
```
hugomontero: Are those lines required for this test?

src/cmd/usb.js
hugomontero: It shouldn't since `finally` will always be executed. 
image

hugomontero (approved):
Just few comments but looks good!

src/cmd/usb-util.js
monkbroc: There's a `device.close()` at end of `executeWithUsbDevice`. This seems extra

monkbroc (approved):
Looks good! I tested a couple commands and it worked well.

PR #757, opened 2024-07-31
Automate Protected Devices to run `particle flash` commands

keeramis:
## Description

Support Protected Devices while running `particle flash xxx` commands

## How to Test

Use `particle flash --local` with devices in DFU mode and normal mode. Test with any platform / any device / any firmware.

**As of this PR**: `develop-6.x` is [this commit](https://github.com/particle-iot/device-os/commit/a6e4efd4be418c2d7156905a6aa054be97349834)

1. If testing Open Devices - please test anything and everything 🙏 (normal mode, dfu mode, any platform, any device-os version)
2. If testing Protected Devices, note that normal mode should work fine. But Protected Devices in DFU mode will pose trouble if using device-os < 6.2.0 (which is the latest `develop-6.x` branch at the time of this PR). Make sure to use `develop-6.x` if you are testing Protected Devices starting in DFU mode. 

❗ Important: Do not run `particle.js update` when your device is on `develop-6.x` branch. This will brick your device. `particle update` obtains the 6.1.1 binary since the reported firmware version from the device is `6.1.1` with the latest develop-6.x and once the bootloader-6.1.1 is flashed on the device with a more advanced system part (develop-6.x system part), it becomes an invalid order and the device will be bricked❗. For a device with `develop-6.x`, the way around this is to flash the 6.1.1 system part first, and then bootloader-6.1.1
To sumarize:
1. If you are on 6.1.1 released firmware and want to go to `develop-6.x` (`6.1.1 --> develop-6.x`)
    1. First, flash the develop-6.x bootloader
    2. Then, flash the develop-6.x system part1
    3. Then, flash the user part
2. If you are on `develop-6.x` firmware and want to go to `6.1.1` (`develop-6.x --> 6.1.1`)
    1. First, flash the 6.1.1 system part1
    2. Then, flash the 6.1.1 bootloader
    3. Then, flash the user part

[Link to binaries for the latest develop-6x branch](https://drive.google.com/drive/folders/1JLQJsI8bbawXvZfZeva1hbdb6WbKQsIW?usp=drive_link)

https://app.shortcut.com/particle/story/129652/automate-protected-devices-to-run-particle-flash-command-e2e-tests
https://app.shortcut.com/particle/story/129651/automate-protected-devices-to-run-particle-flash-command

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing


hugomontero (approved):
LGTM! Jus one quick note: 
Can we put some documentation regarding `executeWithUsbDevice` I think it worth to have in some place that this one opens and close the device so no need to repeat that in the functions that calls this one?
Maybe put as part of the definition here: https://github.com/particle-iot/particle-cli/blob/e58d1bf636998f8157b3a64642d036e48f627ad6/src/cmd/usb-util.js#L78

src/cmd/update.js
```diff
@@ -43,8 +48,10 @@ module.exports = class UpdateCommand extends CLICommandBase {
 		const modulesToFlash = filterModulesToFlash({ modules: deviceOsModules, platformId: device.platformId, allowAll: true });
 		await maintainDeviceProtection({ modules: modulesToFlash, device });
 		const flashSteps = await createFlashSteps({ modules: modulesToFlash, isInDfuMode: device.isInDfuMode , platformId: device.platformId });
-		await flashFiles({ device, flashSteps, ui: this.ui });
+		device = await flashFiles({ device, flashSteps, ui: this.ui });
+
 		this.ui.write('Update success!');
```
monkbroc, line 53: Do you need to reopen the device here like after the other invocations of `flashFiles`

src/cmd/usb-util.js
```diff
@@ -139,36 +141,42 @@ async function executeWithUsbDevice({ args, func, dfuMode = false } = {}) {
...
  * @param {*} deviceId
  * @returns
  */
-async function waitForDeviceToRespond(device) {
+async function waitForDeviceToRespond(device, { timeout = 10000 } = {}) {
 	const deviceId = device.id;
-	const REBOOT_TIME_MSEC = 20000;
-	const REBOOT_INTERVAL_MSEC = 1000;
+	const REBOOT_TIME_MSEC = timeout;
+	const REBOOT_INTERVAL_MSEC = 100;
```
monkbroc, line 155: With this interval reduced, let's pay attention during testing that `waitForDeviceToRespond` doesn't return too early with the old device. For example, if the device got the command to reboot, but hasn't done so after the 100 ms delay, could this return early? I'm not sure if that's a possibility, just pointing out something to keep an eye for. (no change to the code requested here).

monkbroc (approved):
Looks great!

PR #765, opened 2024-08-21
Use the DeviceOS versions endpoints when compiling

suda:
Story details: https://app.shortcut.com/particle/story/129513

### DIFF
--- a/src/cmd/api.js
+++ b/src/cmd/api.js
@@ -154,6 +154,16 @@ module.exports = class ParticleApi {
 		);
 	}
 
+	listDeviceOsVersions(platformId, internalVersion){
+		return this._wrap(
+			this.api.listDeviceOsVersions({
+				platformId,
+				internalVersion,
+				auth: this.accessToken
+			})
+		);
+	}
+
 	compileCode(files, platformId, targetVersion){
 		return this._wrap(
 			this.api.compileCode({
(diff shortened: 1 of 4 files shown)


suda:
@keeramis do you have any insight into why the e2e tests are failing? This seemed to start after the last merge into `main` 🤔

src/cmd/api.js
```diff
@@ -154,6 +154,17 @@ module.exports = class ParticleApi {
 		);
 	}
 
+	listDeviceOsVersions(platformId, internalVersion, perPage=100){
```
hugomontero, line 157: Since we have different params I think might worth to transform it like `{platformId, internalVersion, perPage=100} = {}` unless we can't use the params separately like just send `internalVersion` without `platformId`

hugomontero (approved):
👍

src/cmd/api.js
suda: Good point. In this instance only `platformId` parameter is being used. I kept the positional arguments instead of making them named (with `{}`) to keep the function signature match the others.

keeramis:
@suda It was failing to login. Let's see if rerunning helps. I see that you are doing that already

hugomontero:
> @keeramis do you have any insight into why the e2e tests are failing? This seemed to start after the last merge into `main` 🤔

In this specific case just needed to re-run tests to make it all pass 😄

PR #766, opened 2024-08-21
Handle multi-step flashing on Protected Devices

keeramis:
## Description

This is only applicable for device-os 6.1.1 and lower.

The flashing operation is a multi-step process where each step obtains the device handle, sets the device to the appropriate mode (normal or DFU) based on the binary type, and then flashes the binary onto the device.

Currently, a generic wrapper is used to manage Protected Devices, placing the device in Service Mode only once at the beginning of the flashing process. However, this approach has a limitation: the devices on 6.1.0 and 6.0.0 exit Service Mode and return to a Protected state after flashing the bootloader. You could think of starting in DFU mode, but these device-os versions don't work well on Protected Devices in DFU mode.

The PR throws a downgrade error on Protected Devices if downgrading device-os from 6.1.1 to either 6.1.0 or 6.0.0.

One alternative to downgrade is to convert the Protected Device to Open Device via cloud, and then downgrade via CLI.

## How to Test

1. Flash 6.1.1 on your device. Enable Device Protection on it.
2. Upgrade to develop-6.x or 6.1.1 again -> It will work
3. Downgrade to 6.1.0 again -> It won't work

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/src/cmd/usb-util.js
+++ b/src/cmd/usb-util.js
@@ -167,7 +167,6 @@ async function waitForDeviceToRespond(deviceId, { timeout = 10000 } = {}) {
 	const REBOOT_INTERVAL_MSEC = 500;
 	const start = Date.now();
 	let device;
-
 	while (Date.now() - start < REBOOT_TIME_MSEC) {
 		try {
 			if (device && device.isOpen) {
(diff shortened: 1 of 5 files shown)


src/cmd/usb-util.js
```diff
@@ -183,6 +182,10 @@ async function waitForDeviceToRespond(deviceId, { timeout = 10000 } = {}) {
 			return device;
 		} catch (error) {
 			// ignore errors
+			// device could be open after the last iteration
+			if (device && device.isOpen) {
+				await device.close();
+			}
 		}
```
keeramis, line 189: This is not a part of the PR description. I needed this because I was noticing that if the device never responds to a control request, for example, the UI is hanging there. The UI should return to the next line.

hugomontero (approved):

PR #771, opened 2024-10-07
feat: [sc-131051] Remove access_token query param from CLI

cole-abbeduto-particle:
Story details: https://app.shortcut.com/particle/story/131051

# Story
* Remove query param access_token usage from ApiClient class

## Other
* Cleanup ApiClient so it has consistent handling of auth header/access token
  * Refactor getUser so it doesn't take a token directly anymore. No idea why it ever did
* Refactor WhoAmICommands tests to replace proxyquire with sinon stubs
* Update WhoAmICommands to use api.ensureToken
* Add default null username in settings so we can properly "stub" it in tests
* Change stdout/stderr stubs into spys so they still log for debugging purposes

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -7,6 +7,7 @@ let settings = {
 	apiUrl: 'https://api.particle.io',
 	clientId: 'CLI2',
 	access_token: null,
+	username: null,
 	minimumApiDelay: 500,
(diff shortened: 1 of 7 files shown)


laupow (approved):
LGTM!

PR #780, opened 2024-12-06
Add `particle flash --tachyon`

keeramis:
## Description

Add `particle flash --tachyon` to flash tachyon

Story: https://app.shortcut.com/particle/story/132827/create-particle-flash-tachyon-command

## Usage

1. Checkout this branch. Use `nvm use` to update the node version and then run `npm ci`

One of the following ways to test the command:

-> `particle flash --tachyon /path/to/firmware.zip`

-> `particle flash --tachyon /path/to/firmware/folder`

-> `particle flash --tachyon` from firmware folder

-> `particle flash --tachyon prog_firehose_ddr.elf rawprogram0.xml patch0.xml rawprogram1.xml patch1.xml`

Use `--verbose` to see logs during the update process

## Expected output

See this slack thread - 

[Story](https://app.shortcut.com/particle/story/132827/create-particle-flash-tachyon-command)

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing


src/cli/flash.js
```diff
@@ -37,6 +37,14 @@ module.exports = ({ commandProcessor, root }) => {
 			},
 			'port': {
 				describe: 'Use this serial port instead of auto-detecting. Useful if there are more than 1 connected device. Only available for serial'
+			},
+			'tachyon' : {
+				boolean: true,
+				description: 'Flash Tachyon'
+			},
+			'verbose' : {
```
monkbroc, line 45: You don't need to add `verbose` here. It's a global flag available to all commands already https://github.com/particle-iot/particle-cli/blob/master/src/app/cli.js#L42

src/cmd/flash.js
```diff
@@ -55,11 +59,109 @@ module.exports = class FlashCommand extends CLICommandBase {
...
+		const parsedFiles = await this._analyzeFiles(files);
+		let linxuFiles = await this._findLinuxExecutableFiles(parsedFiles.files, { directory: unpackToolFolder });
+		linxuFiles = linxuFiles.map(f => path.basename(f));
+
+
+		const elfFiles = linxuFiles.filter(f => f.startsWith('prog') && f.endsWith('.elf'));
+		const rawProgramFiles = linxuFiles.filter(f => f.startsWith('rawprogram') && f.endsWith('.xml'));
+		const patchFiles = linxuFiles.filter(f => f.startsWith('patch') && f.endsWith('.xml'));
+
+		if (!elfFiles.length || !rawProgramFiles.length || !patchFiles.length) {
```
monkbroc, line 99: I don't think patch files are mandatory.

src/cmd/flash.js
```diff
@@ -55,11 +59,109 @@ module.exports = class FlashCommand extends CLICommandBase {
...
+			throw new Error('The directory should contain at least one .elf file, one rawprogram file, and one patch file');
+		}
+
+		const sortByNumber = (a, b) => {
```
monkbroc, line 103: I don't think it's necessary to sort the files, as long as the program files are before the patch files.

src/cmd/flash.js
```diff
@@ -55,11 +59,109 @@ module.exports = class FlashCommand extends CLICommandBase {
...
+			const extractNumber = str => parseInt(str.match(/(\d+).xml/)[1]);
+			return extractNumber(a) - extractNumber(b);
+		};
+
+		rawProgramFiles.sort(sortByNumber);
+		patchFiles.sort(sortByNumber);
+
+		if (rawProgramFiles.length !== patchFiles.length) {
```
monkbroc, line 111: I don't think this is necessary either

src/cmd/flash.js
```diff
@@ -55,11 +59,109 @@ module.exports = class FlashCommand extends CLICommandBase {
...
+			throw new Error('The number of rawprogram files should match the number of patch files');
+		}
+
+		let filesToProgram = [];
```
monkbroc, line 115:
```suggestion
		const filesToProgram = [...rawProgramFiles, ...patchFiles];
```

monkbroc (commented):
Looks good :+1: Next step will be to add a version of qdl compiled for each platform.

src/cmd/flash.js
eberseth: I think you want to process in the order given in the command.  The original tools assume that the user correctly orders the raw program and patch files accordingly

keeramis: I brought these back (for time being)

assets/qdl/x64/qdl
monkbroc, line 1: The directory must be `<os>_<architecture>` like `darwin_x64`. Use the same words as in https://binaries.particle.io/particle-cli/manifest.json

src/cmd/flash.js
```diff
@@ -55,11 +59,108 @@ module.exports = class FlashCommand extends CLICommandBase {
...
+			}
+		} else {
+			// If multiple files are passed, check the directory from the first file
+			unpackToolFolder = path.dirname(files[0]);
+		}
+
+		let linxuFiles = await this._findLinuxExecutableFiles(files, { directory: unpackToolFolder });
+		linxuFiles = linxuFiles.map(f => path.basename(f));
+
+		const elfFiles = linxuFiles.filter(f => f.startsWith('prog') && f.endsWith('.elf'));
```
monkbroc, line 93: Change the criteria to
```suggestion
		const elfFiles = linxuFiles.filter(f => f.includes('firehose') && f.endsWith('.elf'));
```
This matches https://www.notion.so/particle/PRD-Particle-CLI-Support-for-Flashing-Tachyon-157d40caa5dc805ab72fdabddf713381?pvs=4#158d40caa5dc80db8131ecd541a48d17

src/cmd/flash.js
monkbroc: I really think this is extra. What if a program doesn't have a patch file to go with it?

src/cmd/flash.js
```diff
@@ -55,11 +59,108 @@ module.exports = class FlashCommand extends CLICommandBase {
...
+
+		this.ui.write('  Patch files:');
+		for (const file of patchFiles) {
+			this.ui.write(`    - ${file}`);
+		}
+		this.ui.write(os.EOL);
+		
+		const archType = utilities.getArchType();
+		const qdl = path.join(__dirname, `../../assets/qdl/${archType}/qdl`);
+		await fs.ensureFile(qdl);
```
monkbroc, line 140: Another piece of feedback: extract the execution of qdl to a separate file `lib/qdl.js`. See the example of the old `dfu.js` https://github.com/particle-iot/particle-cli/blob/v1.43.2/src/lib/dfu.js

Keep using the new execa, but put stuff like getting the correct executable for the current OS and platform in a helper function in `lib/qdl.js` so that `flash.js` just needs to call a function to perform the qdl flash.

eberseth (approved):

src/lib/qdl.js
```diff
@@ -0,0 +1,51 @@
...
+	await fs.chmod(qdlExec, 0o755);
+	return qdlExec;
+}
+
+/**
+ */
+async function run({ files, updateFolder, zip, verbose, ui }) {
+	const qdl = await getExecutable();
+
+	if (zip) {
```
monkbroc, line 24: Make sure that the caller doesn't pass a slash in front of all the files instead of removing it here.

src/lib/qdl.js
```diff
@@ -0,0 +1,51 @@
...
+		updateFolder = updateFolder.replace(/^\//, '');
+		files = files.map((file) => file.replace(/^\//, ''));
+	}
+
+	const qdlArgs = [
+		'--storage', 
+		TACHYON_STORAGE_TYPE,
+		...(zip ? ['--zip', zip] : []),
+		'--include',
+		updateFolder,
```
monkbroc, line 35: Include a conditional here as well in case updateFolder is missing

src/lib/qdl.js
```diff
@@ -0,0 +1,51 @@
...
+		...files
+	];
+
+	ui.write(`Command: ${qdl} ${qdlArgs.join(' ')}`);
```
monkbroc, line 39: Print the command only for verbose mode.

src/cmd/flash.js
```diff
@@ -55,11 +62,136 @@ module.exports = class FlashCommand extends CLICommandBase {
...
+		if (stats.isDirectory()) {
+			updateFolder = path.dirname(input);
+			const dirInfo = await this._extractFlashFilesFromDir(input);
+			includeDir = path.join(path.basename(input), dirInfo.baseDir);
+			filesToProgram = dirInfo.filesToProgram;
+		} else if (utilities.getFilenameExt(input) === '.zip') {
+			updateFolder = path.dirname(input);
+			zipFile = path.basename(input);
+			const zipInfo = await this._extractFlashFilesFromZip(input);
+			includeDir = path.normalize(zipInfo.baseDir);
```
keeramis, line 98: `includeDir = path.normalize(zipInfo.baseDir);`

This gives includeDir as `--include abc/images/qcmxxxx` but the qdl (atleast on mac) needs to have `--include abc/abc/images/qcmxxxx`

This is the folder structure
```
abc.zip
|
|-- abc
|   |--images
|   |     |--qcmxxx
|   |     |   |--prog_firehose_ddr.elf
|   |     |.  |--rawprogram0.elf
```

monkbroc (approved):
:+1: Tested on Mac, Windows and Linux

PR #781, opened 2024-12-10
chore/sc-132686/update-device-constants

hugomontero:
## Description
update device-constants with new platform.

## Discussions
Story details: https://app.shortcut.com/particle/story/132694/create-electron2-platform-in-cloud-services

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     }
   ],
   "dependencies": {
-    "@particle/device-constants": "^3.5.0",
+    "@particle/device-constants": "^3.6.0",
     "binary-version-reader": "^2.5.1",
     "chalk": "^2.4.2",
     "cli-progress": "^3.12.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -52,8 +52,8 @@
     }
   ],
   "dependencies": {
-    "@particle/device-constants": "^3.5.0",
-    "binary-version-reader": "^2.5.1",
+    "@particle/device-constants": "^3.6.0",
+    "binary-version-reader": "^2.5.2",
```
monkbroc, line 56: We shouldn't bump binary-version-reader when updating device-constants. The point of having device-constants as a peer dependency in binary-version-reader is to avoid updating dependencies transiently. Only the top-level application has to update device-constants.

No need to undo it here, but no need to update in other applications.

monkbroc (approved):

hugomontero:
This PR has been included in previous releases

PR #789, opened 2025-01-30
implement cache limit validation

hugomontero:
## Completeness

- [ ] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/src/lib/download-manager.js
+++ b/src/lib/download-manager.js
@@ -59,6 +59,7 @@ class DownloadManager {
 			this.ui.write(`Using cached file: ${cachedFile}`);
 			return cachedFile;
 		}
+		await this._validateCacheLimit({ url, currentFileName: outputFileName });
 		const filePath = await this._downloadFile(url, outputFileName, options);
 		// Validate checksum after download completes
 		// Validate checksum after download completes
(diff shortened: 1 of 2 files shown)


src/lib/download-manager.js
```diff
@@ -219,6 +266,20 @@ class DownloadManager {
...
+
+	async _freeCacheSpace(downloadDirStats) {
+		const { fileStats, totalSize } = downloadDirStats;
+		let deletedSize = 0;
+		for (const file of fileStats) {
+			const { filePath, size } = file;
+			await fs.remove(filePath);
+			this.ui.write(`Removed file: ${file.filePath}`);
+			deletedSize += size;
+			if ((totalSize - deletedSize) <= settings.tachyonCacheLimit) {
```
monkbroc, line 278: Do the units match? totalSize is in bytes and tachyonCacheLimit is in GB.

src/lib/download-manager.js
```diff
@@ -219,6 +266,20 @@ class DownloadManager {
...
 		}
 	}
+
+	async _freeCacheSpace(downloadDirStats) {
+		const { fileStats, totalSize } = downloadDirStats;
+		let deletedSize = 0;
+		for (const file of fileStats) {
+			const { filePath, size } = file;
+			await fs.remove(filePath);
+			this.ui.write(`Removed file: ${file.filePath}`);
```
monkbroc, line 276: No feedback is needed here I think

src/lib/download-manager.js
```diff
@@ -96,6 +97,53 @@ class DownloadManager {
...
+				await this._freeCacheSpace(downloadDirStats);
+			} else {
+				// remove files until the cache limit is satisfied
+				this.ui.write('Cache cleanup skipped. Remove files manually to free up space.');
+			}
+		}
+		return;
+	}
+
+	async getDownloadedFilesStats(currentFile) {
```
monkbroc, line 126: This function can be improved a bit: filter out files that are not normal files (like directories), filter out files that don't end with .zip or .progress in case there's other stuff in there, only call stat once on each file.

monkbroc (changes requested):
Looks good. I suggested some small tweaks

src/lib/download-manager.js
```diff
@@ -96,6 +97,53 @@ class DownloadManager {
...
+		const contentLength = parseInt(fileHeaders.headers.get('content-length') || '0', 10);
+		// get the size of the download directory
+		const downloadDirStats = await this.getDownloadedFilesStats(currentFileName);
+		const totalSize = (downloadDirStats.totalSize + contentLength) / (1024 * 1024 * 1024); // convert to GB
+		if (totalSize > maxCacheSize && downloadDirStats.fileStats.length > 0) {
+			// prompt to delete files
+			const question = {
+				type: 'confirm',
+				name: 'cleanCache',
+				message: 'Downloaded files exceed the cache limit. Do you want to clean up the cache?',
```
monkbroc, line 112: This question should be reworded. The wording of "exceed the cache limit" is unnecessary.
"Do you want to delete previously downloaded versions to free up X GB?"

This means a method to calculate the list of files to be deleted and the size to be deleted needs to be extracted from _freeCacheSpace and called before the question.

src/lib/download-manager.js
```diff
@@ -96,6 +98,73 @@ class DownloadManager {
...
+		const sortedFileStats = fileStats.sort((a, b) => a.mtime - b.mtime);
+
+		return {
+			totalSize: sortedFileStats.reduce((sum, file) => sum + file.size, 0),
+			fileStats
+		};
+	}
+
+	async _shouldCleanCache({ downloadedCacheSizeGB, alwaysCleanCache, maxCacheSizeGB, totalSizeGB, downloadDirStats }) {
+		if (parseInt(maxCacheSizeGB) === 0 || alwaysCleanCache) {
```
monkbroc, line 145: parseInt shouldn't be needed anymore here because it's done where you set `maxCacheSizeGB = parseFloat(settings.tachyonCacheLimitGB)`, right?

monkbroc (approved):
Nice cleanups!

src/lib/download-manager.js
hugomontero: Right! Fixed here! https://github.com/particle-iot/particle-cli/pull/789/commits/e34514f263c69bd0d9e8a44df21ac44fa097235c

PR #791, opened 2025-02-03
Feature/tachyon flash config blob

keeramis:
## Description

- Update QDL binaries with support for `skip-reset` command
- Support `--skip_fashing_os` command but flashing only the config.xml file
- `save_config` will save json file with config params except `registration_code`
- `load_config` loads the config file and adds the then generated `registration_code` to create the config blob
- Retry (3 times) if passwords dont match before quitting the setup process
- Support `~` in ssh path
- Replace`qlril-app` with `particle-ril`

## Completeness

- [ ] Docs have been updated
- [ ] CI is passing


src/cmd/setup-tachyon.js
```diff
@@ -519,13 +552,16 @@ Welcome to the Particle Tachyon setup! This interactive command:
...
 		fs.closeSync(tempFile.fd);
 		return tempFile.path;
 	}
 
-	async _flash(files) {
+	async _flash({ files, skipFlashingOs, output }) {
+
+		const packagePath = files[0];
+
 		const question = {
```
monkbroc, line 565: Will this question be asked if `--load_config` is passed? Remember the goal of load config is to have a setup without a single question. Andrew will use this in the Tachyon CI.

src/cmd/flash.js
```diff
@@ -176,6 +213,20 @@ module.exports = class FlashCommand extends CLICommandBase {
 		return JSON.parse(manifest.toString());
 	}
 
+	async _getFirehoseFileFromZip(zipPath) {
+		const dir = await unzip.Open.file(zipPath);
+		const firehoseFile = dir.files.find(file => file.path.includes('firehose'));
+		if (!firehoseFile) {
+			throw new Error('Unable to find firehose file');
+		}
```
monkbroc, line 221: The logic to find the firehose file is already present in `_extractFlashFilesFromZip` and is more precise: open the zip, load the manifest JSON, get the field for the firehose loader.

Call `_extractFlashFilesFromZip` and get the loader filename from the first element of the returned `files`. Then load that file from the zip.

monkbroc (changes requested):
Looks good with some small tweaks.

src/cmd/setup-tachyon.js
keeramis: This will still ask the question. Let me fix it.

keeramis: I removed this. There is another question about product. Is that OK? or should we get that info from the config file as well?

PR #794, opened 2025-02-18
add region flag in case is not passed use NA

hugomontero:
## Description
Add flag region to be used in tachyon setup.
Region can be stored and read it from the config file.

## How to Test
1. Pull down the branch: `git pull && git checkout feature/sc-133755/add-region-as-setup-option-by-default-use`
2. Attempt to setup a RoW region `npm start -- tachyon setup --region RoW`
3. Attempt to setup a NA (default)  `npm start -- tachyon setup`
4. Attempt to setup and save the config file `npm start -- tachyon setup --save_config test_region.cfg`

**outcome**
* If you pick `RoW` the binary that will be flashed should be `RoW`
* If you don't pick a region, then it will download `NA` by default
* In the config file you'll see a new element named `region`
* In case you don't pass a region as flag it will ask you to pick one.

Story details: https://app.shortcut.com/particle/story/133755

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/setup-tachyon.js
+++ b/src/cmd/setup-tachyon.js
@@ -71,7 +71,6 @@ module.exports = class SetupTachyonCommands extends CLICommandBase {
 					() => this._selectProduct()
 				);
 				config.productId = product;
-				config.region = region;
 			}
 
 			const packagePath = await this._runStepWithTiming(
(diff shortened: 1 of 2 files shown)


src/cmd/setup-tachyon.js
```diff
@@ -477,6 +470,10 @@ Welcome to the Particle Tachyon setup! This interactive command:
 			config.productId = productId;
 		}
 
+		if (region) {
+			config.region = region;
+		}
+
```
monkbroc, line 476: Region doesn't need to go into the config. It's only needed for the download.

hugomontero: Done here! https://github.com/particle-iot/particle-cli/pull/794/commits/ba42e3e8991986611498c9650006493f15e7b38c

PR #797, opened 2025-02-24
Pass initial timezone to Tachyon during setup

monkbroc:
Read time from computer using `Intl.DateTimeFormat().resolvedOptions().timeZone`, or taking from the `--timezone` option, or taking from the JSON config. Pass it to the config blob to Tachyon.

Also tweak a few things in the setup command:
- 3 GB download.
- Put Create new product at the bottom of the list.
- Don't allow blank system password.


hugomontero (approved):
Looks good! 👍

PR #800, opened 2025-02-28
Tachyon setup fix

monkbroc:
Improvements for Tachyon setup:
- Ensure API works after login
- Ensure product name is not blank.
- Ensure SSH key is a public key.
- Don't reference organizations at the top level since not everyone is in an org.


monkbroc (commented):
Verified locally. :shipit:

PR #813, opened 2025-03-27
Use the new document patch format

sergeuz:
Use the new patch format when modifying the device's system doc.


cole-abbeduto-particle (approved):
Perfect! Thank you!

PR #817, opened 2025-04-10
Avoid background update when running update-cli

monkbroc:
Running `particle update-cli` can trigger a background run of `particle update-cli` which leads to a race condition between the 2 update processes.

Rework the background update check to skip an update when running `particle update-cli`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,7 +13,9 @@ module.exports = {
 	},
 	rules: {
 		'no-console': 'off',
-		'valid-jsdoc': 'off'
+		'valid-jsdoc': 'off',
+		'max-depth': ['warn', 5],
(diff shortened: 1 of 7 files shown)


.eslintrc.js
```diff
@@ -13,7 +13,9 @@ module.exports = {
 	},
 	rules: {
 		'no-console': 'off',
-		'valid-jsdoc': 'off'
+		'valid-jsdoc': 'off',
+		'max-depth': ['warn', 5],
+		'max-statements': ['warn', 40]
```
monkbroc, line 18: These get rid of the 29 lint warnings that made it harder to spot the lint errors.

hugomontero (approved):

PR #819, opened 2025-04-16
Support erase tag when flashing Tachyon

monkbroc:
Update QDL tool to support erasing sections while flashing Tachyon.


monkbroc (commented):
Tested by Julien on Linux and Andrew on Mac.

PR #821, opened 2025-04-17
Read GPT tables to find partitions to back up/restore

monkbroc:
Read the GPT partition tables from the Tachyon to find the memory address and size of the partitions to back up and restore.


monkbroc (commented):
Self review. :shipit:

PR #826, opened 2025-04-29
filter out linux devices when opening a device

hugomontero:
## Description
Filter out tachyon devices when flashing firmwares

## How to Test
* Connect a tachyon and another device (the one you want to flash) 
* Run usb list: `npm start -- usb list`
* Run flash --local: `npm start -- flash --local /path/to/flash`

### Outcome Before the change:
* Usb list shows both: tachyon and the other device
* On flashing CLI will ask you to pick the device you want to flash
image

### After the change:
* Usb list still showing both devices.
* On flashing CLI will recognize just one device (not the tachyon) and will start the flashing process right away
image

Story details: https://app.shortcut.com/particle/story/134448/remove-linux-devices-when-flashing-non-linux-devices

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/usb-util.js
+++ b/src/cmd/usb-util.js
@@ -2,7 +2,7 @@ const { spin } = require('../app/ui');
 const { getDevice, isDeviceId } = require('./device-util');
 const { systemSupportsUdev, promptAndInstallUdevRules } = require('./udev');
 const { delay, asyncMapSeries } = require('../lib/utilities');
-const { platformForId } = require('../lib/platform');
+const { platformForId, PLATFORMS } = require('../lib/platform');
 const {
 	getDevices,
 	openDeviceById,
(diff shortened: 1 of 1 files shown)


src/cmd/usb-util.js
```diff
@@ -349,6 +349,10 @@ async function getOneUsbDevice({ idOrName, api, auth, ui, flashMode, platformId
 		devices = devices.filter(d => d.platformId === platformId);
 	}
 
+	// filter out linux kind devices
```
monkbroc, line 352: Is there any command where we'd want getOneUsbDevice to return Linux devices? I don't think so right now, but it would be good to check the list of commands where this function is used.

monkbroc (approved):

src/cmd/usb-util.js
hugomontero: There are some places that currently should not be used for `linux` platforms:
`deviceProtection` (I'm not sure if in a future we're gonna have device protection on tachyons
`esim._changeLed`: Change led (this uses also sendControlRequest so tachyons might not work for this function
`keys.sendPublicKeyToServer`: requires DFU mode to work so `linux` devices might not work for this function
`serial.handleDeviceProtection`: Same here, linux devices might not ready for deviceProtection

PR #833, opened 2025-05-22
Setup slow USB

monkbroc:
Add a USB version line to the device info with a yellow warning (use a USB 3.0 port and USB-C cable for faster flashing) when using a slow cable.

TODO before merging:
- [x] Update particle-usb

### DIFF
--- a/src/cmd/setup-tachyon.js
+++ b/src/cmd/setup-tachyon.js
@@ -70,7 +70,9 @@ module.exports = class SetupTachyonCommands extends CLICommandBase {
 		const requiredFields = ['region', 'version', 'systemPassword', 'productId', 'timezone'];
 		const options = { skipFlashingOs, timezone, loadConfig, saveConfig, region, version, variant, board, skipCli };
 		await this.ui.write(showWelcomeMessage(this.ui));
-		this.deviceId = await this._verifyDeviceInEDLMode();
+		const { deviceId, usbVersion } = await this._verifyDeviceInEDLMode();
+		this.deviceId = deviceId;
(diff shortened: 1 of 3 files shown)


src/cmd/setup-tachyon.js
```diff
@@ -147,8 +153,8 @@ module.exports = class SetupTachyonCommands extends CLICommandBase {
 				ui: this.ui,
 			}));
 		} catch (error) {
-			// ignore error and return default values
-			this.ui.write('We couldn\'t get the device info.');
+			// If this fails, the flash won't work so abort early.
+			throw new Error('Unable to get device info. Please restart the device and try again.');
```
monkbroc, line 157: If I Ctrl-C the flash and start it again, I get the warning here then printing the device info crashes with "cannot read properties of undefined reading deviceId" from _printDeviceInfo so I think the error is fatal.

src/cmd/setup-tachyon.js
```diff
@@ -475,7 +504,7 @@ module.exports = class SetupTachyonCommands extends CLICommandBase {
 		// Display the formatted step
 		this.ui.write(`${os.EOL}===================================================================================${os.EOL}`);
 		this.ui.write(`Step ${step}:${os.EOL}`);
-		this.ui.write(`${text}${os.EOL}`);
+		this.ui.write(`${text}`);
```
monkbroc, line 507: This tightens the output.

hugomontero:
image
Great warning message!
image

hugomontero (approved):
Looks great!

PR #842, opened 2025-06-25
remove board in config file to prevent board overwriting

hugomontero:
## Description
Remove `board` from config in order to prevent board type overwriting.
`--board` flag still present in case is necessary.

## How to Test
1. Attempt to setup a device with a config file: `npm start tachyon setup -- --load_config ${config-file}
The config file should have `board` property (e.g.):
```
{
  "region": "NA",
  "version": "latest",
  "variant": "desktop",
  "board": "formfactor_dvt",
  "skipCli": false,
  "systemPassword": "${encrypted_password}",
  "productId": 9999,
  "timezone": "America/Merida",
  "wifi": {
    "ssid": "{SSID}",
    "password": "{PASSWORD}"
  },
  "country": "US"
}
```

**outcome**
Even with `board` in config, the device should use the right `board` according device information.

story details: https://app.shortcut.com/particle/story/134831/remove-board-from-config-tachyon-setup

## Completeness

- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/src/cmd/setup-tachyon.js
+++ b/src/cmd/setup-tachyon.js
@@ -230,6 +230,8 @@ module.exports = class SetupTachyonCommands extends CLICommandBase {
 			try {
 				const data = fs.readFileSync(loadConfig, 'utf8');
 				const config = JSON.parse(data);
+				// remove board to prevent overwriting.
+				delete config.board;
 				return { ...config, silent: true, loadedFromFile: true };
 			} catch (error) {
 				throw new Error(`The configuration file is not a valid JSON file: ${error.message}`);
@@ -741,7 +743,6 @@ module.exports = class SetupTachyonCommands extends CLICommandBase {
 		const configFields = [
 			'region',
 			'version',
-			'board',
 			'variant',
 			'skipCli',
 			'systemPassword',


technobly (approved):
I used an old config with board = formfactor, and got the correct dvt-formfactor download.  I added a little line just to see when it reads in the file.  I'm guessing it doesn't overwrite the bad file and remove .board, but that's also probably fine.  The system still uses .board and likely writes it out to new files as well... which then would be ignored on next use.  Maybe we should remove it before writing out to configs as well?

```
 -  Device ID: 422a060000000000a20a7d31
 -  Region: NA
 -  OS Version: Ubuntu 20.04
 -  USB Version: 3.0
loaded config! ===============================================
Skipping to Step 5 - Using configuration file: /Users/brett/code/tachyon-config-125.json 


===================================================================================

Step 7:

Next, we'll download the Tachyon Operating System image.
Heads up: it's a large file — 3GB! Don't worry, though—the download will resume
if it's interrupted. If you have to kill the CLI, it will pick up where it left. You can also
just let it run in the background. We'll wait for you to be ready when its time to flash the device.
Using cached file: /Users/brett/.particle/downloads/tachyon-ubuntu-20.04-NA-headless-formfactor_dvt-1.0.130.zip
```

PR #847, opened 2025-07-17
feat: [sc-134988] Support app push/remove/list with applications as arrays

cole-abbeduto-particle:
Story details: https://app.shortcut.com/particle/story/134988

# Changes
* Add check for migrated applications before push/remove/list commands so the user knows they need to update their particle-linux version
* Update push to use upsert action
* Update remove to use remove action
* Update list to support rendering array style apps/containers

### DIFF
--- a/src/cmd/app.js
+++ b/src/cmd/app.js
@@ -291,7 +291,7 @@ module.exports = class AppCommands extends CLICommandBase {
 			}
 
 			// We can assume at this point that apps is an array since we migrate both at the same time on device and we return above
-			const apps = deviceDoc.features?.applications?.properties?.apps;
+			const apps = _.get(deviceDoc, 'features.applications.properties.apps');
 			if (!apps || apps.length === 0) {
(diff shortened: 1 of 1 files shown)


src/cmd/app.js
```diff
@@ -237,24 +239,22 @@ module.exports = class AppCommands extends CLICommandBase {
...
-			});
-
 			// Use PATCH method to update the device document
 			await this.api.patchDocument({
 				productId: device.product_id,
 				deviceId: device.id,
 				docName: 'system',
-				patchOps
+				patchOps: {
+					action: 'upsert',
```
monkbroc, line 250: This is interesting. Should the Console use action: 'upsert' when adding Wi-Fi networks by SSID as well?

monkbroc (approved):
Looks good in principle. Will test with particle-linux

src/cmd/app.js
cole-abbeduto-particle: Yes, I just added upsert support in the sentinel PR and I let @CarlosHdez know about it the other day. We'll get it in console eventually

PR #865, opened 2025-09-02
Setup 20.04 by default, and allow setup of 24.04

monkbroc:
## Description

Ensure that when setting up Tachyon, we default to using Ubuntu 20.04. Allow setting up 24.04 with `--distro_version 24.04`

## How to Test

Currently there are no released 24.04 images, but we need this PR before we do to ensure we don't start flashing 24.04 accidentally.

So this is expected to fail:
```
particle tachyon setup --distro_version 24.04

Step 7 failed with the following error: No build available for the provided parameters
```

This should still work:
```
particle tachyon setup
```

### DIFF
--- a/src/cli/tachyon.js
+++ b/src/cli/tachyon.js
@@ -28,6 +28,9 @@ module.exports = ({ commandProcessor, root }) => {
 			board: {
 				description: 'Board to download package for'
 			},
+			distro_version: {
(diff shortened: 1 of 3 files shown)


monkbroc:
Thanks Andrew. I confirmed by pasting the JSON into the CLI code and verifying that it does pick 20.04 by default and can be overridden with the option.

PR #867, opened 2025-09-29
Feature/sc 135566/update restore command

hugomontero:
## Completeness

- [ ] Contributor has signed [CLA](https://part.cl/icla)
- [ ] Docs have been updated
- [ ] CI is passing

### DIFF
--- a/src/cmd/backup-restore-tachyon.js
+++ b/src/cmd/backup-restore-tachyon.js
@@ -4,6 +4,7 @@ const path = require('path');
 const fs = require('fs-extra');
 const os = require('os');
 const temp = require('temp').track();
+const unzip = require('unzipper');
 const {
 	addLogHeaders,
 	getEDLDevice,
@@ -11,7 +12,9 @@ const {
 	prepareFlashFiles, handleFlashError
 } = require('../lib/tachyon-utils');
 const settings = require('../../settings');
-const { compressDir } = require('../lib/utilities');
+const { compressDir, fileExists, delay } = require('../lib/utilities');
+const ParticleApi = require('./api');
+const createApiCache = require('../lib/api-cache');
 
 const PARTITIONS_TO_BACKUP = ['nvdata1', 'nvdata2', 'fsc', 'fsg', 'modemst1', 'modemst2'];
 
@@ -21,6 +24,7 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 		this.ui = ui || this.ui;
 		this._baseDir = settings.ensureFolder();
 		this._logsDir = path.join(this._baseDir, 'logs');
+		this._setupApi();
 	}
 
 	async backup({ 'output-dir': outputDir = process.cwd(), 'log-dir': logDir = this._logsDir, existingLog } = {}) {
@@ -96,6 +100,8 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 	async restore({
 		'input-dir': inputDir = process.cwd(),
 		'log-dir': logDir = this._logsDir,
+		'force-cloud': forceCloud,
+		filePath,
 		existingLog
 	} = {})	{
 		const device = await getEDLDevice({ ui: this.ui });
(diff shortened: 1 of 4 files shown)


src/cmd/backup-restore-tachyon.js
```diff
@@ -152,4 +166,66 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 		}
 	}
 
+	async _getFilePathToRestore({ forceCloud, filePath, deviceId, inputDir }) {
+		const defaultFileName = path.join(inputDir, `manufacturing_backup_${deviceId}.zip`);
+		const defaultFileExists = await fileExists(defaultFileName);
+		if (filePath) {
+			const exists = await fileExists(filePath);
+			console.log(exists);
```
laupow, line 174: Did you mean to leave the `console.log` in?

src/cmd/backup-restore-tachyon.js
```diff
@@ -152,4 +166,66 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
...
+			.promise();
+
+		// Decide what to return
+		if (roots.size === 1) {
+			const [root] = Array.from(roots);
+			const resolved = path.join(tmpOutputDir, root);
+			return resolved;
+		} else {
+			return tmpOutputDir;
+		}
```
monkbroc, line 216: What does this do?

src/cmd/backup-restore-tachyon.js
```diff
@@ -96,6 +100,8 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 	async restore({
 		'input-dir': inputDir = process.cwd(),
 		'log-dir': logDir = this._logsDir,
+		'force-cloud': forceCloud,
+		filePath,
```
monkbroc, line 104: the command line argument is lowercase `filepath`

src/cmd/backup-restore-tachyon.js
```diff
@@ -152,4 +166,66 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
...
+		const defaultFileExists = await fileExists(defaultFileName);
+		if (filePath) {
+			const exists = await fileExists(filePath);
+			console.log(exists);
+			if (await fileExists(filePath)) {
+				this.ui.stdout.write(`Using zip file from ${filePath} ${os.EOL}`);
+				return filePath;
+			}
+			this.ui.stdout.write(`Unable to find file at ${filePath} ${os.EOL}`);
+			throw new Error('Unable to find file at ' + filePath);
```
monkbroc, line 180: Don't both write and throw an error because the error will be printed in red already by the CLI when it exits.

I tested it and the output is not normal for the CLI. If an error is thrown, it should print a message in red (with a stack trace with `--verbose`). That tells me something is catching errors and reprinting them. That's extra code that's not needed, unless there's a very specific output expected by this command.

```
Restoring NV data to device 422a060000000000d939257c...
Logs will be saved to /home/monkbroc/.particle/logs/tachyon_422a060000000000d939257c_restore_1759263660412.log
false
Unable to find file at foobar 
An error ocurred while trying to restore up your tachyon 
Error: Unable to find file at foobar 
Verify your logs /home/monkbroc/.particle/logs/tachyon_422a060000000000d939257c_restore_1759263660412.log for more information 
```

monkbroc (changes requested):
Looks good. Needs a couple of tweaks.

src/cmd/backup-restore-tachyon.js
hugomontero: When I uncompress on my mac the zip file that uncompressed into a dir hehe. So I was preparing this in case the files are into a dir. After that I recall Mac always uncompress the dir and doesn't mean the zip actually contains a folder inside. I'll remove that part to just return the tmpOutputDir hehe.

src/cmd/backup-restore-tachyon.js
hugomontero: That's right I was catching that error, but I remove that part here https://github.com/particle-iot/particle-cli/pull/867/commits/d3719484072d800d64334c4a96c378d6839fd861 since actually it was exiting as succeeded instead of errored.

src/cmd/backup-restore-tachyon.js
hugomontero: fixed here! https://github.com/particle-iot/particle-cli/pull/867/commits/d3719484072d800d64334c4a96c378d6839fd861

PR #922, opened 2026-05-22
Chore/sc 137859/standardize api calls

hugomontero:
## Description
* Standardize the way we use api calls. 
* Make sure all auth checks are centralized and the error message are the same for every command
* Make sure VErrors are returning the trace when we use --verbose option

## How to Test
* Run thru every command with a valid and invalid token and check the output

**Outcome**
* Commands should work with tokens
* Without a valid toke it should print that the user requires to be logged in.
* Commands that doesn't require token should still working even without token

Story details: https://app.shortcut.com/particle/story/137859/there-are-some-optional-cloud-calls-that-prevent-the-user-work-when-they-are-not-logged-in

## Completeness

- [x] Contributor has signed [CLA](https://part.cl/icla)
- [x] Docs have been updated
- [x] CI is passing


monkbroc (approved):
👍 Reviewed with Claude and by scanning the changes. Tested some commands locally and everything looks good.

PR #925, opened 2026-06-12
[A] fix installer to validate correctly the checksum

hugomontero:
## Description
Fix unix installer for particle-cli sha-256 check.
*  Eyeballing the installer script at installer/unix/install-cli shows that you're using 'shasum a 256... ' instead of 'shasum -a 256... ' on line:
* Also the sha was checking the uncompressed file instead of the compressed one/

## How to Test
* Try to use the sh file to install the cli.

**outcome**
* Sha verification should happen
* it will validate the compressed file instead of the uncompressed one.

Issue: https://github.com/particle-iot/particle-cli/issues/909

## Completeness

- [x] Contributor has signed [CLA](https://part.cl/icla)
- [x] Docs have been updated
- [x] CI is passing

### DIFF
--- a/installer/unix/install-cli
+++ b/installer/unix/install-cli
@@ -90,13 +90,19 @@ mkdir -p "$DEST_PATH"
 TMP_FILE=$(mktemp)
 
 echo ":::: Downloading CLI version $VERSION"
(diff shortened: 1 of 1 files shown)


monkbroc (approved):
Tested locally 👍

PR #930, opened 2026-08-18
Add timestamp option to particle serial monitor

monkbroc:
Add `--timestamp` and `--utc` options to `particle serial monitor` to help line up logs between the device and cloud.


hugomontero (approved):
Tested locally both: `--timestamp` `--timestamp --utc` works good!

Issue #76, opened 2015-07-31
`particle serial info`


kennethlimcp:
Adding a command to retrieve the system information for the Photon and the Electron will be easily than getting people to open up a serial terminal and hit "s"

zsup:
Agreed; would also be helpful to take the opportunity to make that info more human readable

kennethlimcp:
definitely. :+1:

emilyrose:
with the Electron it sounds like we'll also need to handle parsing additional information like IMEI/SIM card info.

PR #933, opened 2026-08-30
tachyon: make setup recover from unfamiliar GPTs and fix 24.04 onboarding

mrlambchop:
## What

Nine commits that make `particle tachyon setup` work on 24.04, and make
`tachyon backup` / `tachyon restore` layout-independent and verified.

| commit | change |
|---|---|
| `3acc9124` | write the config blob to the **post-flash** `misc`; verify NV restores |
| `da0ccaa0` | log the real UFS LUN geometry read back from the device GPT |
| `d8431859` | identify a device whose layout has no `boot_a`/`boot_b` |
| `e19bb122` | offer the headless variant on 24.04, drop the beta label |
| `9dccf641` | fetch eSIM profiles during 24.04 setup, as 20.04 already does |
| `54bd8d90` | tell 24.04 users the modem will activate, as 20.04 does |
| `ba0dea1f` | print the durable backup path, not only its basename |
| `7111fea6` | make the outgoing GPT advisory for setup; EDL identity is sufficient |
| `640ece58` | resolve region/board by explicit input → device → cloud → prompt; remove silent NA/DVT defaults |

No dependency changes. `package.json` is untouched — in particular this does **not**
add `@particle/tachyon-image`, which stays dormant until the A/B work lands.

### 1. The config blob was written using the wrong GPT

`createConfigBlobStep` resolved the `misc` LBA from the **pre-flash** GPT and
`flashOSAndConfigStep` programmed that XML after the OS was written. On a 20.04 → 24.04
setup that means the blob went to the old `misc` offset — i.e. to whatever the new image
put on those sectors. On a device already running 24.04 it failed outright with
`Partition misc not found in device partition table`, because 24.04 had no `misc` at all
(fixed by tachyon-composer #76).

Reordered to:

1. create the blob (`createBlobFile`, unchanged);
2. **preflight the image** — `verifyConfigPartitionStep` reads the partition table out of
   the image about to be flashed and fails if it has no `misc`, if `misc` is smaller than
   256 sectors, or if the blob would not fit. This runs *before* the multi-GB write, so a
   mismatch costs seconds rather than a full flash;
3. flash the OS/GPT, keeping the device in EDL;
4. read the newly flashed live GPT off the device;
5. resolve `misc` **by label** and generate the QDL program XML.

`flash()` already ran `flashTachyon` then `flashTachyonXml` as two qdl invocations, so
step 4 slots between them. `createConfigBlobStep` now returns only `{ configBlobPath }`.
Step 5 (`resolveConfigPartition`) is the authoritative gate: if `misc` is absent from the
post-flash GPT it fails with a message naming the image and stating that the OS was
installed but the configuration was not applied, rather than silently writing the blob to
whatever now occupies those sectors. `keepInEdl = skipReset || Boolean(configBlobPath)`.

Reset behaviour is unchanged from `master`: `flashTachyonXml({ skipReset })` is called
exactly as before. Only the internal name changed (`shouldResetOS` → `keepInEdl`).

### 2. `restore()`'s retry path called `backup()`

`backup-restore-tachyon.js:151-153` caught a restore failure and retried with
`this.backup({ 'input-dir': … })` — wrong method, and `backup()` does not take
`input-dir`. Now retries `this.restore(...)` with restore's own arguments.

Also: backups stored durably under the Particle data directory with the path always
printed; a `backup.json` sidecar (schema version, device ID, source LUN/LBA geometry,
byte lengths, SHA-256 per file); pre-write validation that each file fits its target
partition; and post-write read-back hash verification. Legacy archives without
`backup.json` are still accepted.

### 3. `identify` threw on any device without `boot_a`

`partitionDefinitions` threw `Partition boot_a not found in device partition table` for
any layout lacking it. 20.04 has `boot_a`/`boot_b`; 24.04 does not — so `tachyon setup`
died at "Unable to get device info." on a 24.04 board. Adds an `optionalPartitions` list
(threaded through `prepareFlashFiles`) so a missing optional partition yields `null` and
is filtered out instead of throwing, and uses the absence of both boot partitions
together with a single non-A/B `system` to report `Ubuntu 24.04`.

### 4. 24.04 offered no headless variant

The `ubuntu24` workflow had a desktop variant only, so step 10 printed
`undefinedLearn more about Tachyon…` — a missing `setupCompletedMessage`. Adds the
`headless` variant with its own completion message, and drops `(beta)` from the 24.04
name plus the now-stale `selectionWarning`.

### 5. 24.04 setup never fetched the eSIM profiles

The `ubuntu24` workflow was missing **`getCountryStep`** and **`getESIMProfilesStep`**.
`ubuntu20` has both. So 24.04 setup never asked for a country, never called
`PUT /v1/products/:productId/devices/:deviceId/target_profile`, and wrote the config blob
with **no `esim` key at all**.

The consumer has been there the whole time — particle-linux's bootstrap declares

```ts
esim?: { profiles: Array<{ iccid, activation_code }>, target_profile: string | null }
```

and calls `storeEsimBootstrap()` whenever the blob carries it. 24.04 devices simply never
got one, so a device worked only if its eUICC happened to already be provisioned.

Both steps are OS-agnostic: `getCountryStep` prompts (or takes `country` from a loaded
config in silent mode), and `getESIMProfilesStep` is non-fatal — it warns and returns
`{ esim: null }` if the lookup fails. `createBlobFile` snapshots the whole context minus a
small exclusion list and drops nulls, so `country` and `esim` reach the blob with no extra
plumbing.

Ordering matters and is now asserted: the country must resolve before the profile lookup
(it is the `countryCode` argument), and both before the blob is built, since the blob is a
snapshot of the context.

### 6. 24.04's completion message did not mention the modem

Both 24.04 variants listed "Connect to the Particle Cloud" but not "Activate the built-in
5G modem"; 20.04 lists both. That was accurate while `ubuntu24` had no
`getESIMProfilesStep`, and stopped being accurate the moment it gained one. Added to the
desktop and headless variants, matching `ubuntu20`'s wording and punctuation.

Tied to behaviour rather than left as prose: a test asserts a workflow claims modem
activation **if and only if** its step list contains `getESIMProfilesStep`, across every
variant — so the copy and the capability cannot drift apart again the way they just did.

> ⚠️ **Not verified end to end on hardware.** head2's eUICC is already provisioned, so
> re-running setup there does not exercise a device that arrives without a profile —
> which is exactly the case this message now promises. Needs an unprovisioned unit before
> release.

### 7. The outgoing GPT is no longer a setup precondition

Commits `7111fea6` and `640ece58` fix the underlying recovery contract rather than
adding another known layout to the allow-list.

#### Why setup read the old GPT in the first place

The original structure appears to have been a convenience: `getTachyonInfo()`
produced one object containing the EDL device ID plus values inferred from storage:

- region from a modem marker in `fsg`;
- EVT/DVT board type from the LUN containing `nvdata1`;
- installed OS from `boot_a`/`boot_b`;
- whether manufacturing data looked present.

Region and board then selected the matching image automatically, while the remaining
fields were printed to the user. That is useful metadata, but setup treated failure to
obtain any of it as proof that flashing could not work. The conclusion does not follow:
the EDL enumeration already supplies the identity used by setup, and the selected image
carries `prog_firehose_ddr.elf`, its rawprogram XMLs, patch XMLs, and Primary/Backup
GPT writes.

#### New behavior

The Particle Cloud lookup starts as soon as USB enumeration supplies the EDL device ID,
in parallel with the slower best-effort storage read. A blank, corrupt, or unfamiliar GPT
now emits a warning and setup continues. Region and board are resolved independently, in
this precedence:

1. an explicit command-line option;
2. a loaded setup configuration;
3. a value read from the physical device;
4. Particle Cloud metadata;
5. an interactive prompt.

The physical device wins over the cloud because the cloud record is only as fresh as its
last connection. Today the cloud supplies region from `modem_firmware_version`; it does
not report EVT/DVT board type, so an unreadable board still prompts unless the operator
provided `--board` or saved it in the setup configuration. The selected source is shown
in the device information. There is no silent NA/DVT fallback.

#### Effect on the onboarding flow

1. Login and EDL enumeration run as before; EDL supplies the device ID without reading storage.
2. The cloud lookup starts immediately from that ID while the existing GPT read runs. Cloud
   failure is non-fatal.
3. Setup resolves region and board separately using the precedence above. A source may answer
   one field without being trusted for the other.
4. The resolved values and any non-device provenance are printed before image selection. If a
   value remains unknown, setup asks the operator instead of guessing.
5. The normal variant, account, product, Wi-Fi, eSIM, image preflight, flash, post-flash GPT
   read, and `misc` configuration steps then continue unchanged.

A real `TachyonConnectionError` is deliberately different: setup still offers a retry
and stops if the user declines, because the device is no longer satisfying the one
hardware precondition -- being reachable in EDL.

#### Downsides and remaining constraints

- The cloud record may be stale, so a readable physical value deliberately wins. If the
  device cannot answer, the output labels cloud-derived values.
- A loaded config is now treated as explicit intent, including `board`, and overrides
  discovery. Reusing a config created for different hardware can therefore select the wrong
  image; explicit `--region` / `--board` remain the final override.
- When no source knows a value, interactive setup pauses for a region or board choice. This
  is intentional: guessing NA/DVT could install an image for the wrong cellular bands or
  hardware revision.
- Keeping identification as best-effort adds one firehose/read attempt before a recovery
  flash. Removing it entirely would be faster, but would regress automatic image
  selection for the normal readable-device case.
- EDL presence is the only required *device state*, not a guarantee that every image is
  compatible. The package must still contain a valid programmer/GPT and fit the target
  UFS geometry.
- The post-flash `misc` read remains intentional. It occurs only after the image has
  installed its own GPT and resolves where the configuration blob belongs. If that new
  image lacks a valid `misc`, the OS is installed but setup reports that configuration
  could not be applied.

The setup tests cover recognized and unsupported layouts, cloud lookup ordering, every
precedence tier, prompt fallback, loaded board configuration, connection failure, and
connection retry. Utility tests cover cloud firmware parsing, non-fatal lookup failure,
and refusing to classify an unfamiliar GPT as DVT.

#### Hardware regression: 24.04 single-slot to 20.04 A/B

Tested on head2, device `422a060000000000d0c7965f`, using the exact command:

```console
pi@head2-pi:~/images $ /opt/particle/bin/particle tachyon setup --version /home/pi/images/tachyon-ubuntu-20.04-NA-headless-formfactor_dvt-1.0.190.zip
```

Before installing the PR build, the old `/opt` binary stopped before any write:

```text
Starting Process. See logs at: .../tachyon_flash_422a060000000000d0c7965f_1788376808082.log
Tachyon setup failed: Unable to get device info. Please restart the device and try again.
```

After installing the PR build at the same path, the command:

- passed identification from the outgoing 24.04 single-slot layout;
- preflighted the supplied image's `misc`;
- wrote the supplied 20.04 A/B GPT and OS;
- re-read that new GPT;
- programmed the configuration into `misc`;
- completed all 12 setup steps.

The booted board reports Ubuntu 20.04.6 LTS, kernel `5.4.219`, root
`/dev/sda11`, Wi-Fi associated, and the Particle, GNSS, and RIL services active.
`misc` resolves to `/dev/sda5`, is exactly 1 MiB, and hashes as an all-zero 1 MiB
buffer after first boot, showing that bootstrap consumed and erased the configuration.

The older 20.04 image reports systemd `degraded`; observed failures across these boots
were `persist.mount`, `sfsconfig.service`, and `thermal-engine.service`. Those boot-time
image issues were recorded but are outside this CLI change.

#### Hardware verification of the final precedence resolver

The `087b8c8d` ARM64 build exercised the final behavior at `/opt/particle/bin/particle`. The final `640ece58` adds only the numbered precedence comment in source; its exact ARM64 artifact was built and installed afterward. The same
20.04 package was run again, with the saved setup configuration streamed after removing
its `region` and `board` keys so discovery—not the config—had to resolve both fields.

The live output was:

```text
 -  Device ID: 422a060000000000d0c7965f
 -  Board: DVT or later
 -  Region: NA (from Particle Cloud)
 -  OS Version: Ubuntu 20.04
```

This exercises the per-field merge rather than an all-or-nothing fallback: the physical
GPT supplied DVT, while the physical `fsg` partition was readable but contained no
`SG560D-NA` / `SG560D-EM` region marker, so the cloud filled only region. `fsg` itself
and its manufacturing-data `EFS` header are present; this is not modem-data damage.

Setup again completed all 12 steps. The board booted Ubuntu 20.04.6, associated to Wi-Fi,
and the Particle, GNSS, and RIL services are active. The post-flash `misc` is 1 MiB and
hashes as an all-zero buffer after bootstrap consumed the configuration.

## Tests

1117 passing, 10 pending, lint clean. New coverage in
`steps.test.js`, `tachyon-utils.test.js` and `backup-restore-tachyon.test.js` for: the
GPT being read *after* the OS flash and the new `misc` LBA being used; an image with no
`misc`, or an undersized one, failing at preflight before any flash; a device with no
`misc` post-flash failing at the second gate; restore retrying `restore`; restore
rejecting a file that would overrun its target; read-back hash verification; legacy
archives with no `backup.json`; and `partitionDefinitions` tolerating absent optional
partitions.

## Hardware

head2, against composer build `1.2.14-dev+build.64a0cda` (tachyon-composer #76). The
board has been on 24.04 since 27 Aug, so **the first two runs are 24.04 → 24.04** — which is the
case defect 1 broke outright. The 20.04 → 24.04 direction is not exercised here.

### Run 1

- setup ran to step 10 with the headless completion message rendered, no `undefined`
- device boots to a login on `Ubuntu 24.04.4 LTS`, kernel `6.8.0-1058-particle`
- `misc -> sda3`, exactly 1048576 bytes, and **entirely zero** after first boot — the blob
  was written to the correct post-flash offset, consumed by the particle-linux bootstrap
  and wiped
- the configured password authenticates (sha512crypt match for `particle` in `/etc/shadow`)
- Wi-Fi associated to the configured SSID and took a DHCP lease
- modem NV byte-identical to the pre-flash capture across all six partitions
  (`modemst1` `bf3b6f09…`, `modemst2` `a00c952e…`, `fsc` `fa43239b…`, `fsg` `abccf416…`,
  `nvdata1`/`nvdata2` `106f0647…`), with `nvdata1/2` still at the 20.04 1.5 MiB sizes

### Run 2 — repeat setup, again on 24.04

Both runs exercise the case that used to fail outright with
`Partition boot_a not found in device partition table` → "Unable to get device info."
That failure is itself the evidence the board was on 24.04: the very first attempt in
this session died there, because 24.04 has no `boot_a`.

- `Device info:` now reports **`OS Version: Ubuntu 24.04`** and setup continues
- **Step 8 — "Checking that the operating system image can store your configuration…"** —
  the new image preflight ran and passed before any flashing
- ran through to step 10, device boots to a login again on `root=PARTLABEL=system` / `sda4`
- `misc` again exactly 1 MiB and again entirely zero after first boot
- Wi-Fi reassociated, same DHCP lease
- modem identity intact — `particle-tachyon-ril-ctl info` reports IMEI-1 `8644300100010 91`,
  IMEI-2 `…92`, ICCID `8935711800001245077`, EID `89044045…`, firmware `SG560DNAPAR60A03`

`fsg`, `fsc`, `nvdata1` and `nvdata2` are byte-identical across **both** flashes.
`modemst1`/`modemst2` differ between the two runs, which is expected and not
flash-related — they are the modem's live working EFS partitions. Verified directly:
with no flashing at all, `modemst2` changed hash within 90 s of idle runtime
(`f05c5fc1…` → `99ff9042…`). After the first flash they matched the pre-flash capture
exactly, which is what shows the flash itself does not touch them.

### On-image suites (run 2 device, `TMPDIR=/tmp`)

| suite | result | baseline |
|---|---|---|
| `particle-linux-test` | 21 passed / 0 failed / 2 skipped | 20 / 0 / 3 |
| `particle-tachyon-gnss-test` | 41 passed / 2 failed / 9 skipped | 43 / 0 / 9 |
| `particle-tachyon-rild-test` | 74 passed / 3 failed | 78 / 0 / 0 |

The failures are pre-existing image issues, not regressions from this PR — see the
composer PR for the reasoning. Note the suites must be run with a valid `TMPDIR`: under
`adb shell`, `TMPDIR` defaults to the Android path `/data/local/tmp`, which does not
exist on this image, and the suites' output redirects fail in a way that looks like a
CLI fault (`'particlectl systemdoc' exited 2`).

## Landing note

Land **after** tachyon-composer #76 and after an image containing `misc` is published —
this PR's post-flash gate fails by design against an image that has no `misc`.

### DIFF
--- a/src/cli/tachyon.js
+++ b/src/cli/tachyon.js
@@ -32,7 +32,8 @@ module.exports = ({ commandProcessor, root }) => {
 				description: 'Board to download package for'
 			},
 			distro_version: {
-				description: 'Linux distribution version to use'
+				description: 'Linux distribution version to use',
+				type: 'string'
 			},
 			skip_cli: {
 				description: 'Do not log in the Particle CLI',
@@ -67,7 +68,8 @@ module.exports = ({ commandProcessor, root }) => {
 				type: 'string'
 			},
 			distro_version: {
-				description: 'Linux distribution version to use'
+				description: 'Linux distribution version to use',
+				type: 'string'
 			}
 		},
 		handler: (args) => {
@@ -178,4 +180,3 @@ module.exports = ({ commandProcessor, root }) => {
 
 	return tachyon;
 };
-
--- a/src/cli/tachyon.js
+++ b/src/cli/tachyon.js
@@ -116,7 +116,7 @@ module.exports = ({ commandProcessor, root }) => {
 	commandProcessor.createCommand(tachyon, 'backup', 'Backup Tachyon NV data', {
 		options: {
 			'output-dir': {
-				description: 'Directory to save the backup files'
+				description: 'Directory to save the backup files. Defaults to the backups folder in the Particle data directory'
 			},
 			'log-dir': {
 				description: 'Directory to save the log files'
@@ -136,7 +136,7 @@ module.exports = ({ commandProcessor, root }) => {
 	commandProcessor.createCommand(tachyon, 'restore', 'Restore Tachyon NV data', {
 		options: {
 			'input-dir': {
-				description: 'Directory containing the NV data files'
+				description: 'Directory containing the NV data files. Defaults to the backups folder in the Particle data directory'
 			},
 			'log-dir': {
 				description: 'Directory to save the log files'
--- a/src/cmd/backup-restore-tachyon.js
+++ b/src/cmd/backup-restore-tachyon.js
@@ -14,9 +14,19 @@ const {
 } = require('../lib/tachyon-utils');
 const settings = require('../../settings');
 const VError = require('verror');
-const { compressDir, fileExists } = require('../lib/utilities');
+const { compressDir, fileExists, sha256File } = require('../lib/utilities');
+const { createHash } = require('crypto');
 
+// Modem NV. `persist` is deliberately absent: the WLAN and BT MACs live in modem
+// NV (read over AT with +QNVR), not there, and /persist holds only rmtfs symlinks
+// that the device recreates -- and reformats -- on its own if they go missing.
 const PARTITIONS_TO_BACKUP = ['nvdata1', 'nvdata2', 'fsc', 'fsg', 'modemst1', 'modemst2'];
+
+// Sidecar describing what was captured, so a restore can verify the bytes instead
+// of trusting six unlabelled blobs. Archives without it still restore.
+const BACKUP_METADATA_FILE = 'backup.json';
+const BACKUP_SCHEMA_VERSION = 1;
+const SECTOR_SIZE_IN_BYTES = 4096;
 
 module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 	constructor({ ui } = {}) {
@@ -24,10 +34,14 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 		this.ui = ui || this.ui;
 		this._baseDir = settings.ensureFolder();
 		this._logsDir = path.join(this._baseDir, 'logs');
+		// Default to the Particle data directory rather than the working
+		// directory: modem NV is not reproducible, and a backup left in whatever
+		// folder the command happened to run from is one `cd` away from being lost.
+		this._backupDir = path.join(this._baseDir, 'backups');
 		this._setupApi();
 	}
 
-	async backup({ 'output-dir': outputDir = process.cwd(), 'log-dir': logDir = this._logsDir, existingLog } = {}) {
+	async backup({ 'output-dir': outputDir = this._backupDir, 'log-dir': logDir = this._logsDir, existingLog } = {}) {
 		const device = await getEDLDevice({ ui: this.ui });
 		const outputDirExist = await fs.exists(outputDir);
 		const logDirExist = await fs.exists(logDir);
@@ -46,13 +60,18 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 		addLogHeaders({ outputLog, startTime, deviceId: device.id, commandName: 'Tachyon backup' });
 		try {
 			const tmpOutputDir = await temp.mkdir('tachyon_backup');
+			let sourcePartitions = [];
 			const { firehosePath, xmlFile } = await prepareFlashFiles({
 				logFile: outputLog,
 				ui: this.ui,
 				partitionsList: PARTITIONS_TO_BACKUP,
 				dir: tmpOutputDir,
 				device,
-				operation: 'read'
+				operation: 'read',
+				modifyPartitions: (partitions) => {
+					sourcePartitions = partitions;
+					return partitions;
+				}
 			});
 			const files = [
 				firehosePath, // must be first
@@ -68,6 +87,7 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 				serialNumber: device.serialNumber
 			});
 			await qdl.run();
+			await this._writeBackupMetadata({ dir: tmpOutputDir, deviceId: device.id, partitions: sourcePartitions });
 			// zip file
 			const compressedFile = await compressDir({
 				pathToCompress: tmpOutputDir,
@@ -80,6 +100,9 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 			fs.appendFileSync(outputLog, `SHA256: ${compressedFile.sha256}${os.EOL}`);
 			this.ui.stdout.write(`Created File: ${compressedFile.outputFile}${os.EOL}`);
 			this.ui.stdout.write(`Backing up NV data from device ${device.id} complete!${os.EOL}`);
+			this.ui.stdout.write(
+				`Restore it with: particle tachyon restore --filepath ${compressedFile.outputFile}${os.EOL}`
+			);
 		} catch (error) {
 			const { retry } = await handleFlashError({ error, ui: this.ui });
 			if (retry) {
@@ -98,7 +121,7 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 	}
 
 	async restore({
-		'input-dir': inputDir = process.cwd(),
+		'input-dir': inputDir = this._backupDir,
 		'log-dir': logDir = this._logsDir,
 		filepath,
 		existingLog
@@ -114,14 +137,16 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 		this.ui.stdout.write(`Restoring NV data to device ${device.id}...${os.EOL}`);
 		this.ui.stdout.write(`Logs will be saved to ${outputLog}${os.EOL}`);
 		addLogHeaders({ outputLog, startTime, deviceId: device.id, commandName: 'Tachyon restore' });
+		let zipFilePath;
 		try {
-			const zipFilePath = await this._getFilePathToRestore({
+			zipFilePath = await this._getFilePathToRestore({
 				filepath,
 				deviceId: device.id,
 				inputDir
 			});
 			const tempPath = await this.extractZipFile(zipFilePath);
 			// check the file unzip it in a temp and use it to prepare and flash.
+			let targetPartitions = [];
 			const { firehosePath, xmlFile } = await prepareFlashFiles({
 				logFile: outputLog,
 				ui: this.ui,
@@ -129,8 +154,17 @@ module.exports = class BackupRestoreTachyonCommand extends CLICommandBase {
 				dir: tempPath,
 				device,
 				operation: 'program',
-				checkFiles: true
+				checkFiles: true,
+				modifyPartitions: (partitions) => {
+					targetPartitions = partitions;
+					return partitions;
+				}
 			});
+			// Everything that can be checked without touching the device is checked
+			// before the first byte is written: a half-written NV set is worse than
+			// a refused restore.
+			const metadata = await this._readBackupMetadata({ dir: tempPath, deviceId: device.id });
+			await this._verifyBackupFiles({ partitions: targetPartitions, metadata });
 			const files = [
 				firehosePath, // must be first
 				xmlFile,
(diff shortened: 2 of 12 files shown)


monkbroc:
Note that there's no headless variant for latest for 24.04

```
? Select the OS Type to setup in your device Ubuntu 24.04

===================================================================================

Step 1:

Select the variant of the Tachyon operating system to set up.
The 'desktop' includes a GUI and is best for interacting with the device with a keyboard, mouse, and display.
The 'headless' variant is accessed only by a terminal out of the box. 

? Select the OS variant: Headless (command-line only)
Tachyon setup failed: No builds found for this variant headless, board formfactor_dvt and version latest
```

monkbroc:
Added DO-NOT-MERGE to make it clear that this needs to wait until the changes to Ubuntu 24.04 in Tachyon composer are finalized.

mrlambchop:
### Additional head2 hardware matrix — 2026-09-03

Tested device `422a060000000000d0c7965f` with the final `640ece58` ARM64 CLI installed at `/opt/particle/bin/particle`.

| Transition | Setup result | Boot result |
|---|---|---|
| Ubuntu 24.04/1.2.23 single-slot → Ubuntu 20.04/1.0.191 headless | Passed all 12 steps; outgoing 24.04 GPT identified without requiring `boot_a`; DVT came from the physical GPT and NA from Particle Cloud | Passed: Ubuntu 20.04.6, kernel `5.4.219`, root `/dev/sda11` (`system_a`); Particle/GNSS/RIL active; 1 MiB `misc` consumed and zeroed |
| Ubuntu 20.04/1.0.191 → Ubuntu 24.04/1.1.43 desktop | Passed all 12 steps, including the restored eSIM step and image/config preflight | **Failed in the 1.1.43 image:** U-Boot repeatedly reports `Unexpected EFS image header magic ... Failed to setup` and never reaches Linux |
| Freshly flashed 1.1.43 disk layout → Ubuntu 24.04/1.2.23 desktop | Passed all 12 steps; setup identified the 1.1.43 GPT, ran eSIM retrieval, preflighted `misc`, and replaced the A/B-style layout with 1.2.23 single-slot `system` | Passed: Ubuntu 24.04.4, kernel `6.8.0-1058-particle`, root `/dev/sda4`; Wi-Fi connected; Particle/GNSS/RIL active; 1 MiB `misc` consumed and zeroed; UART reached the Linux login prompt |

The reported 1.1.43 → 1.2.23 boot stop did **not** reproduce on this device with the fixed workflow. The runtime log confirms that the newly restored onboarding step made it into the blob and was consumed:

```text
INFO: RilService: Found esim bootstrap. Applying, this may take a while.
INFO: RilService: Finished esim bootstrap
```

The 1.0.191 → 1.1.43 failure is not an NV overwrite by setup. `fsg` is byte-identical in backups from Aug 26, Aug 30, and after the failed boot: SHA-256 `abccf416cf41458e37ca0f02e9ad604523a6af71b9fab23d26dd8a5b7223a0b3`. Version 1.1.43 U-Boot rejects this existing EFS format; 1.0.191 and 1.2.23 both boot with it. This also exposes a separate diagnostic weakness: `tachyon identify` reports `Manufacturing data: Found` because it currently searches for the bytes `EFS` anywhere in `fsg`, which is less strict than the bootloader header validation.

Release/cache notes from the run:

- `tachyon-1.1.43.json` is not published. The public `stable` manifest selects the exact 1.1.43 NA desktop artifact; that is how the 1.1.43 leg was pinned and verified before writing.
- Non-interactive setup aborts if the 10 GB download-cache limit requires cleanup because it attempts to prompt. Cached ZIPs were moved aside, not deleted, for this test.
- The Pi rebooted once during the 1.2.23 download. No device write had begun; the resumable download continued correctly after the host returned and Tachyon was placed back in EDL.

mrlambchop:
### Why the reported 1.1.43 → 1.2.23 update stopped in UEFI

The two package GPTs make the corruption path explicit:

- 1.1.43 `misc`: LUN 0, sector `392`, byte offset `0x188000`, 256 sectors / 1 MiB
- 1.2.23 `efi`: LUN 0, sectors `6..131077`, starting at `0x6000`
- 1.2.23 `misc`: LUN 0, sector `138758`, byte offset `0x21e06000`, 256 sectors / 1 MiB

The previous setup flow resolved `misc` from the outgoing 1.1.43 GPT, flashed the 1.2.23 GPT/image, then wrote the config blob at stale sector 392. Sector 392 is inside the new 1.2.23 `efi` partition, so that final 1 MiB configuration write corrupts the EFI image. This directly explains a successful flash followed by a pre-kernel UEFI failure.

This PR preflights `misc` from the target image, then re-reads the live GPT after the OS flash and writes at the new sector 138758. The hardware run above booted through to Linux using that sequence.

The missing 24.04 `getESIMProfilesStep` shown elsewhere in this PR is a separate onboarding bug. It affects the config consumed later by `particle-linux`; by itself it cannot explain a failure before Linux boots.

mrlambchop:
### Exact 1.0.191 → 1.2.23 hardware result

Follow-up tested the exact pair on head2 with final CLI commit `640ece58`:

1. Flashed and boot-verified Ubuntu 20.04/1.0.191 headless: Ubuntu 20.04.6, kernel `5.4.219`, root `/dev/sda11` / `system_a`, Particle/GNSS/RIL active, and `misc` consumed/zeroed.
2. Re-entered EDL and ran setup with `--version 1.2.23 --distro_version 24.04`, selecting the NA desktop build.
3. Setup identified the outgoing OS as Ubuntu 20.04, ran all 12 steps including eSIM retrieval and target-image `misc` preflight, flashed the single-slot GPT/image, re-read that GPT, and wrote configuration to the new `misc`.
4. Boot-verified Ubuntu 24.04.4, kernel `6.8.0-1058-particle`, root `/dev/sda4` / `system`, Wi-Fi connected, Particle/GNSS/RIL active, and the 1 MiB `misc` consumed/zeroed.

**Result: 1.0.191 → 1.2.23 works with this PR. It does not work safely with the old CLI.**

1.0.191 places `misc` at LUN 0 sector `392` (`0x188000`). In 1.2.23, EFI spans sectors `6..131077` and the new `misc` is sector `138758` (`0x21e06000`). The old CLI retained sector 392 from the outgoing GPT and wrote the config blob there after flashing 1.2.23, overwriting 1 MiB inside EFI. It can therefore report setup success and then leave the board stuck in UEFI. The post-flash GPT lookup in this PR is required for this transition.

mrlambchop:
Follow-up: wire `--distro_version` into setup workflow selection

The option was originally added in `3575c6e6` and directly filtered manifest builds in the old monolithic setup implementation. Refactor `0eb720d5` moved manifest filtering to `workflow.osInfo`, but did not use `distroVersion` to select that workflow. The setup flag consequently became a no-op while the same flag continued to work in `tachyon download-package`.

New setup rules:

1. A local ZIP is authoritative; its embedded manifest selects the workflow. A conflicting `--distro_version` now fails clearly.
2. For remote builds, explicit `--distro_version 20.04` or `--distro_version 24.04` selects that Ubuntu workflow and skips the OS question.
3. An explicit distro overrides a workflow stored in `--load_config`, consistent with normal CLI-over-config precedence.
4. Without the flag, existing behavior remains: use the loaded workflow, otherwise ask interactively, with the silent default remaining Ubuntu 20.04.
5. Unsupported Linux distro versions fail immediately and list the supported versions.

The Ubuntu 20.04 and 24.04 workflows continue to have the same onboarding step sequence; only their OS metadata, variants, and completion copy differ. Runtime `distroVersion` is now synchronized with the selected workflow as well.

Validation: lint passed; all 1,121 unit tests passed; all 10 integration tests passed. Added focused coverage for both 20.04 and 24.04 prompt skipping, interactive metadata synchronization, CLI-over-config precedence, unsupported versions, and local-image mismatch rejection.

PR #934, opened 2026-08-31
tachyon: identify from the EDL device ID and the cloud, not partition sniffing

mrlambchop:
## The problem

`particle tachyon identify` derived everything it printed from a partition read — region by scanning `fsg` for `SG560D-NA` / `SG560D-EM`, OS version by sniffing markers in `boot_a` / `boot_b`. It had no other source, so when the sniff came up empty the information was simply lost.

Two concrete failures on a 24.04 build (`422a060000000000d0c7965f`):

- **`fsg` holds no `SG560D` string at all.** Not a hyphen mismatch — `strings fsg.backup | grep SG560D` returns nothing on this unit. So `identify` printed `Region: Unknown` while the device's own `/opt/particle/syscon.json` said `"region": "NA"`.
- **The 24.04 layout has no `boot_a`**, so `identify` failed outright with `Partition boot_a not found in device partition table` — *after* it had already read the device ID successfully.

## The approach

A device in EDL mode reports its full device ID over plain USB enumeration, before any firehose upload or partition read:

```
> getEdlDevices()
{ serialNumber: 'D0C7965F', id: '422a060000000000d0c7965f' }
```

That ID is enough to ask the cloud what the device is, so ask there first. The cloud records the region in `modem_firmware_version` — the same `SG560D` + `NA`/`EM` string the manufacturing test parses ([`read-region.sh`](https://github.com/particle-iot/tachyon-overlays/blob/main/overlays/manufacturing/files/read-region.sh)) — along with the serial number, device name, product, and last-reported image.

The on-device read still runs. It is the only source of the manufacturing-data check, and the only source at all when the machine is offline or the CLI is not logged in. What changed is that it can no longer sink the command.

**Precedence:** the device wins on anything it could actually answer — that is the hardware in front of you, whereas a cloud record is only as fresh as the last handshake. The cloud fills gaps and adds what it alone knows. The output says when a value came from the cloud. Fields nothing could supply are omitted rather than printed as `Unknown`.

The lookup never throws: not logged in, device never connected, and no network all leave the local answer untouched, so `identify` keeps working offline and unauthenticated as it does today.

## Why it matters beyond cosmetics

Setup treats an `Unknown` region as `NA`:

```js
optionsFromDevice.region = deviceInfo.region.toLowerCase() !== 'unknown' ? deviceInfo.region : 'NA';
```

So a RoW device whose marker is missing silently gets an NA image — different cellular bands, and a device that cannot reach its network.

## Verified on hardware (head2, real EDL device)

**On this branch alone (based on `master`, so the 24.04 partition read still fails):**

```
Could not read the device directly: Partition boot_a not found in device partition table
Verify your logs /tmp/tachyon_422a060000000000d0c7965f_identify_1788145638626.log for more information

Device ID: 422a060000000000d0c7965f
Name: bobs-yellow-hat-d0c7965f
Region: NA (from Particle Cloud)
Serial number: P06400000000000
Modem firmware: SG560DNAPAR60A01
Image: 1.2.17 (headless)
Product: 35245
```

Previously this run produced nothing but the `boot_a` error. `Manufacturing data` and `OS Version` are correctly absent — only the local read can supply them.

**Stacked on #933, where the local read succeeds:**

```
Device ID: 422a060000000000d0c7965f
Name: bobs-yellow-hat-d0c7965f
Region: NA (from Particle Cloud)
Serial number: P06400000000000
Modem firmware: SG560DNAPAR60A01
Manufacturing data: Found
OS Version: Ubuntu 24.04
Image: 1.2.17 (headless)
Product: 35245
```

Both sources merged. `Region: NA` corroborates the device's own `syscon.json`, and `Serial number: P06400000000000` matches it exactly.

## Notes for review

- **Stacking:** this is based on `master` and will conflict with #933 in `src/lib/tachyon-utils.js` (both add to the block above `module.exports` and to the export list). The conflict is additive and mechanical. #933 should land first — without it, `identify` cannot complete a local read on 24.04 at all.
- Tests live in a new `src/cmd/identify-tachyon.test.js` rather than `src/lib/tachyon-utils.test.js`, which #933 also creates, to avoid an add/add conflict. Happy to move them once #933 lands.
- 17 new tests; full unit suite 1076 passing, lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SSL8pbMoVJzoAkb6uXtkZA

### DIFF
--- a/src/cmd/identify-tachyon.js
+++ b/src/cmd/identify-tachyon.js
@@ -5,38 +5,107 @@ const os = require('os');
 
 const {
 	getEDLDevice,
-	getTachyonInfo, handleFlashError
+	getTachyonInfo, handleFlashError,
+	lookupCloudDeviceInfo
 } = require('../lib/tachyon-utils');
 
 
 module.exports = class IdentifyTachyonCommand extends CLICommandBase {
-	constructor({ ui } = {}) {
+	constructor({ ui, api } = {}) {
 		super();
 		this.ui = ui || this.ui;
+		this.api = api || this._particleApi().api;
 	}
 
+	/**
+	 * A device in EDL mode reports its device ID over USB enumeration, before any
+	 * firehose upload or partition read. Everything else worth knowing hangs off
+	 * that ID in the cloud, so ask for it there first.
+	 *
+	 * The on-device read still runs, because it is the only source of the
+	 * manufacturing-data check and the only source at all when the machine is
+	 * offline or the CLI is not logged in. It is no longer allowed to sink the
+	 * command: a partition layout this version of the CLI cannot parse used to make
+	 * `identify` fail outright, even though the device ID had already been read.
+	 */
 	async identify() {
 		const device = await getEDLDevice({ ui: this.ui });
 		const outputLog = path.join(os.tmpdir(), `tachyon_${device.id}_identify_${Date.now()}.log`);
 
+		const cloudInfo = await lookupCloudDeviceInfo({ deviceId: device.id, api: this.api });
+		const localInfo = await this._readFromDevice({ device, outputLog });
+
+		if (!cloudInfo && !localInfo) {
+			this.ui.stdout.write(`Device ID: ${device.id}${os.EOL}`);
+			this.ui.stdout.write(`Could not read this device, and it is not in your Particle account.${os.EOL}`);
+			this.ui.stdout.write(`Verify your logs ${outputLog} for more information${os.EOL}`);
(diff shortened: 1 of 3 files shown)


monkbroc (approved):
Tested with 24.04 and 20.04 NA Tachyon. I don't have a RoW Tachyon to try.

hugomontero (approved):
Code looks good!

PR #937, opened 2026-09-15
update device-constants to 4.2.0

XuGuohui:
Add new platform `Sulu`.


XuGuohui:
The `Sulu` platform is marked as `public:false` in device-constants, causing some of the tests failed. How can we use the Particle CLI with Sulu to do some tasks during development, e.g. flashing local firmwares, identify/inspecting devices?

monkbroc:
Hugo updated device-constants to make Sulu public to avoid issues.

